PR #5, opened 2024-03-15
Base Encoder, Decoder, Field Maps

jonathanmcdermid:
Create base classes for encoder and body decoder, change field encoding, decoding to use maps so functionality can be overridden.


rldoris:
Looks good. Approved.

rldoris (approved):
Looks good. Approved

PR #7, opened 2024-03-18
Conan

jonathanmcdermid:
Use conan package manager to install dependencies.
Reverts #37


CMakeLists.txt
```diff
@@ -133,6 +128,29 @@ if(GIT_EXECUTABLE)
...
+find_package(spdlog_setup REQUIRED)
+find_package(spdlog REQUIRED)
+find_package(fmt REQUIRED)
+find_package(cpptoml REQUIRED)
+
+include_directories(${nlohmann_json_INCLUDE_DIR})
+include_directories(${spdlog_INCLUDE_DIR})
+include_directories(${spdlog_setup_INCLUDE_DIR})
+include_directories(${fmt_INCLUDE_DIR})
+include_directories(${cpptoml_INCLUDE_DIR})
```
valgur, line 145: Prefer using `target_link_libraries` with the CMake targets provided by the package config files loaded using `find_package()`. They provide and propagate the necessary linked libraries, include dirs, defines and build flags via `INTERFACE_INCLUDE_DIRECTORIES` and similar properties applied to the targets.

valgur: Also, unlike `target_include_directories`, the global `include_directories()` info won't get exported with the generated CMake config files that should eventually be installed by the package.

CMakeLists.txt
```diff
@@ -133,6 +128,29 @@ if(GIT_EXECUTABLE)
...
+
+set(CONAN_LIBS 
+    nlohmann_json::nlohmann_json 
+    spdlog::spdlog
+    spdlog_setup::spdlog_setup
+    fmt::fmt
+)
```
valgur, line 152: While not wrong for a library consisting of only a small number of components, it's usually preferable to do `target_link_libraries(<target> PUBLIC/PRIVATE/INTERFACE nlohmann_json::nlohmann_json ...)` to apply the dependency info directly to the relevant target (which is `common` in this case) and have CMake propagate the dependency automatically. PUBLIC/PRIVATE should be based on whether the dependency is used in the public headers of the component or not.

conanfile.txt
```diff
@@ -0,0 +1,10 @@
+[requires]
+nlohmann_json/3.11.2
+gegles-spdlog_setup/1.1.0
+cpptoml/0.1.1
+spdlog/1.13.0
+fmt/10.2.1
```
valgur, line 6: `fmt` and `cpptoml` are transitive dependencies and don't really need to be explicitly listed, unless you want to specifically constrain their versions for some reason. This makes version conflicts with ConanCenter transitive dependencies more likely, though, unless you also create a lock file to pin the packages to specific revision. Conan v2 is not very forgiving about version conflicts, unfortunately, which is why opted to use version ranges in my PR.

conanfile.txt
```diff
@@ -0,0 +1,10 @@
...
+
```
valgur, line 7:
```suggestion

[layout]
cmake_layout

```
Makes it easier to `conan install .`. Otherwise you need to provide additional arguments for the build dir location, etc.

CMakeLists.txt
```diff
@@ -133,6 +128,29 @@ if(GIT_EXECUTABLE)
...
 
+add_compile_definitions(SPDLOG_FMT_EXTERNAL)
+add_compile_definitions(FMT_HEADER_ONLY)
+add_compile_definitions(GTEST_USE_STATIC_LIBS)
+
+find_package(nlohmann_json REQUIRED)
+find_package(spdlog_setup REQUIRED)
+find_package(spdlog REQUIRED)
+find_package(fmt REQUIRED)
+find_package(cpptoml REQUIRED)
```
valgur, line 139: These are probably unnecessary as they get loaded automatically as transitive dependencies of `spdlog` and `spdlog_setup`.

conanfile.txt
valgur, line 7:
```suggestion

[test_requires]
gtest/1.14.0

```
To get rid of the FetchContent for GTest.

conanfile.txt
```diff
@@ -0,0 +1,10 @@
...
+[generators]
+CMakeDeps
+CMakeToolchain
```
valgur, line 10:
```suggestion
CMakeToolchain

[options]
spdlog/*:shared=True
```
spdlog does not really handle static linkage very well. Its logger registry is implemented as a global singleton, which gets erroneously re-initialized in each shared library and/or executable that has statically linked against it.

.github/workflows/build.yml
```diff
@@ -39,18 +39,37 @@ jobs:
...
+
       - name: Install CMake
         uses: lukka/get-cmake@latest
         
       - name: Configure
         shell: bash
         run: |
           if [ "${{ env.COMPILER }}" == "msvc" ]; then
-            cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DCMAKE_CXX_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_C_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc
+            cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DCMAKE_CXX_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_C_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_MODULE_PATH=build/$BUILD_TYPE/generators -DCMAKE_PREFIX_PATH=build/$BUILD_TYPE/generators
```
valgur, line 68:
```suggestion
            cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DCMAKE_CXX_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_C_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_TOOLCHAIN_FILE=build/$BUILD_TYPE/generators/conan_toolchain.cmake
```
This is the way Conan expects the generated dependencies to be used typically. Are you specifically trying to avoid this approach for some reason?

CMakeLists.txt
jonathanmcdermid: done!

.github/workflows/build.yml
valgur: Ok, makes sense. That's annoying indeed.

Another option is to use the generated CMake presets instead.
See https://github.com/valgur/velodyne_decoder/blob/master/.github/workflows/build.yml#L101-L149 for an example.

Alternatively, you can let the official conan_provider.cmake module (that I've made use of in my PR) take care of creating a matching Conan configuration, installation and use of Conan dependencies automatically.

CMakeLists.txt
jonathanmcdermid: done!

CMakeLists.txt
jonathanmcdermid: done!

conanfile.txt
jonathanmcdermid: done!

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.14)
```
valgur, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```
Conan v2 requires CMake 3.15 or greater anyway.

.github/workflows/build.yml
```diff
@@ -39,33 +39,48 @@ jobs:
...
+            elif [ "${{ env.COMPILER }}" == "gcc" ]; then
+              cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G Ninja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_CXX_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-gnu -DCMAKE_C_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-gnu -DCMAKE_TOOLCHAIN_FILE=build/$BUILD_TYPE/generators/conan_toolchain.cmake
+            fi
+          elif [ "${{ env.OPERATING_SYSTEM }}" == "windows" ]; then
+            if [ "${{ env.COMPILER }}" == "msvc" ]; then
+              cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DCMAKE_CXX_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_C_COMPILER_TARGET=x86_64-pc-$OPERATING_SYSTEM-msvc -DCMAKE_TOOLCHAIN_FILE=build/generators/conan_toolchain.cmake
+            elif [ "${{ env.COMPILER }}" == "clang-cl" ]; then
+              cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G Ninja -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_MODULE_PATH=build/generators/ -DCMAKE_PREFIX_PATH=build/generators/
```
valgur, line 71:
```suggestion
              cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -T ClangCL
```
The ClangCL compiler is a part of MSVC, so it needs a VS project generator together with `ClangCL` MSVC toolset to be set to work correctly. See https://stackoverflow.com/questions/64189111/how-to-use-clang-cl-llvm-with-vs2019-cmake

.github/workflows/build.yml
```diff
@@ -39,33 +39,48 @@ jobs:
...
       - name: Test
-        # TODO: figure out why unit testing doesn't work on windows clang
-        if: |
-            !(env.OPERATING_SYSTEM == 'windows' && env.COMPILER == 'clang')
         run: ctest --test-dir build --build-config ${{ env.BUILD_TYPE }} --output-on-failure
 
       - name: Regression
-        # TODO: figure out why regression testing doesn't work on windows gnu or windows clang
-        if: |
-            !(env.OPERATING_SYSTEM == 'windows' && (env.COMPILER == 'gcc' || env.COMPILER == 'clang'))
```
valgur, line 68: MinGW GCC will fail to find its C++ runtime DLL when running any novatel_edie executables unless it is explicitly handled. I've done so here: https://github.com/novatel/novatel_edie/blob/d15f0b7bd714c5d3868fac390448be95fbbbb18c/CMakeLists.txt#L125-L139

jonathanmcdermid: thanks!

valgur:
@jonathanmcdermid With #37 merged, you should run `git add . --renormalize` and commit the changes before merging the changes from `main` to avoid merge conflicts in existing branches.

rldoris (approved):

profiles/linux/Debug/clang
valgur, line 1: A separate profile file for Release and Debug is a bit of an overkill, imo. You can specify the build type with `-s build_type=...` when doing `conan install`.

src/hw_interface/stream_interface/src/filestream.cpp
```diff
@@ -102,14 +102,14 @@ void FileStream::OpenFile(FILE_MODES eMode)
 void FileStream::CloseFile()
 {
     MyStream.close();
-    if (MyStream.fail()) { throw NExcept("\"%ls\" close file failed", clFilePath.u32string().c_str()); }
```
valgur, line 105: This accidentally reverts merged changes.

profiles/linux/Debug/clang
jonathanmcdermid: Good point!

PR #8, opened 2024-03-25
clang-tidy

jonathanmcdermid:
make use of clang-tidy


src/decoders/novatel/api/fileparser.hpp
```diff
@@ -75,14 +75,14 @@ class FileParser
     //
     //! \param[in] sDbPath_ Filepath to a JSON message DB.
     //----------------------------------------------------------------------------
-    FileParser(const std::string sDbPath_);
+    FileParser(std::string sDbPath_);
```
rldoris, line 78: Why wouldn't we want this to be const? It looks like it's feeding to the Parser which takes in a const `Parser(const Parser&)`

jonathanmcdermid: this is better addressed in the resharper MR, but const in a declaration wont actually have any effect so its generally recommended to not use it. The exception to this is when the variable is a pointer or reference, (which sDBPath is in the resharper mr, but not here).

rldoris (approved):

PR #11, opened 2024-04-12
resharper

jonathanmcdermid:
This is just me applying as many of the resharper suggestions I can stomach, the size of the MR is primarily due to how much changes when you enforce common naming conventions. There really shouldn't be any functional changes, so if you see something that looks like one, point it out and we can discuss.


src/decoders/novatel/src/rangecmp/range_decompressor.cpp
```diff
@@ -6,7 +32,7 @@ using namespace novatel::edie;
 using namespace novatel::edie::oem;
 
 //-----------------------------------------------------------------------
-//! Table used to expand the scaled Pseudorange STDs.  This is defined
+//! Table used to expand the scaled pseudo range STDs.  This is defined
```
zainarnoldhexagon, line 35: not super important, but I think pseudorange is the correct term here.

zainarnoldhexagon (commented):
I didn't see anything out of the ordinary! I would approve, but I don't know if that auto-merges your changes, so when you've got eyes on this from other people and you're ready, just ping me!

src/decoders/novatel/src/rangecmp/range_decompressor.cpp
jonathanmcdermid: fixed!

Brian-Pho:
Good changes here

PR #22, opened 2024-05-03
Use Conan or Vcpkg to manage third-party dependencies

valgur:
Hi!

This PR replaces all of the vendored libraries under lib/ with integrated Conan support to manage the dependencies and an additional optional Vcpkg manifest, which can alternatively be used as well.

There are a few benefits to using a dedicated package manager instead of vendoring the libraries:
- Prevents possible multiple-definition issues and version conflicts for common libraries like `nlohmann-json` and `spdlog` for the users of novatel_edie who rely on these as well.
- Provides more flexibility for the user to select whether the third-party libs should be linked as static, shared or header-only libs.
- Easier for the user to update the dependencies, if necessary. To provide fixes for security vulnerabilities, for example.
- Makes the library much easier to package.

Some limitations of this approach:
- Requires Python with Conan installed to be present.
- Automatic use of Conan in CMake requires CMake v3.24+. There are hacky workarounds for this, but I did not want to go there.
- Should add a lock file to pin the package versions and revisions, since the ConanCenter packages are updated from time-to-time to bump dependency versions, for example, which can sometimes break the build.

The included conanfile.py does a bit more than the CMakeLists.txt is currently capable of in this PR. I left the installation of the novatel_edie libraries and third-party libs for a separate later PR.

I also tweaked the root CMakeLists.txt a bit to get rid of some artificial limitations that did not seem justified to me - building on unexpected OS-s, with unexpected compilers (such as AppleClang) or without Git being present. `-Werror` is generally better left for development and CI builds as well, since it can easily break the build for unexpected warnings added by newer compiler versions.


valgur:
I would like to merge this one before opening a PR for the Ptyhon bindings. Would cause too many conflicts otherwise.

jonathanmcdermid:
I'm new to vcpkg and conan, and I also have an [open PR](https://github.com/novatel/novatel_edie/pull/7) that attempts to install the dependencies we currently house within the lib folder. I was trying to get it to work with just a conanfile.txt, but I was running into issues with gtest. Are you able to accomplish what you've done in this PR with a conanfile.txt instead of a .py?

valgur:
Yes, the conanfile.txt you used is more or less sufficient. Only a couple of tweaks are needed that I added as comments under that PR.

conanfile.py becomes necessary when you want to provide support for `conan create .` to create a Conan package from the project. It also provides finer control over option handling and validation. This becomes more relevant with the Python bindings, where I don't want to use shared linkage with spdlog (which is a reasonable default otherwise) and statically link everything instead to keep Python packaging simpler.

valgur:
@jonathanmcdermid I fixed all of the CI failures and the PR should be ready to be merged as far as I'm concerned. I had to disable the MinGW compilers because the shared build of spdlog seems to have issues with exporting symbols correctly when built using these compilers. I can look into it into a follow-up PR.

.github/workflows/build.yml
```diff
@@ -16,10 +16,10 @@ jobs:
...
           - [windows, windows-latest, msvc, Debug]
           - [windows, windows-latest, msvc, Release]
-          - [windows, windows-latest, gcc, Debug]
-          - [windows, windows-latest, gcc, Release]
-          - [windows, windows-latest, clang, Debug]
-          - [windows, windows-latest, clang, Release]
+          # - [windows, windows-latest, gcc, Debug]
+          # - [windows, windows-latest, gcc, Release]
+          # - [windows, windows-latest, clang, Debug]
+          # - [windows, windows-latest, clang, Release]
```
jonathanmcdermid, line 22: are you able to get these configurations to work with conan?

valgur:
@jonathanmcdermid I fixed the issues with MinGW GCC - the tests failed due to a missing C++ runtime DLL - and replaced MinGW Clang with ClangCL. The latter is better supported on Windows and probably the more popular choice as well.

jonathanmcdermid:
Sorry it took me a while to get to this! I need to update my github notifications so these dont go under the radar for me. I appreciate you simplifying the conanfile. I'll be working my way through this PR this week.

.github/workflows/build.yml
```diff
@@ -43,56 +43,33 @@ jobs:
...
-              fi
-          done
-          
-          if [ ${#failures[@]} -gt 0 ]; then
-              echo "One or more regression tests failed! ${failures[@]}"
-              echo "See the README if you don't know how to fix this."
-              exit 1
-          fi
-          
+          bash regression/run_tests.sh ./build/bin/$BUILD_TYPE/converter_components
```
jonathanmcdermid, line 72: So is the advantage of moving the regression testing script to a shell script that we can run it locally? If so, the README should be updated to reflect this. I didnt notice any functional changes otherwise.

CMakeLists.txt
```diff
@@ -91,20 +101,24 @@ include_directories(
...
+    add_subdirectory(src/decoders/novatel/test)
+    add_subdirectory(src/hw_interface/stream_interface/test)
+endif()
+
+if(BUILD_EXAMPLES)
+    add_subdirectory(examples)
+endif()
 
 if(WIN32)
+    # Copy third-party library DLLs to for tests
```
jonathanmcdermid, line 120: rephrase this

examples/novatel/command_encoding/command_encoding.cpp
```diff
@@ -25,10 +25,12 @@
...
 
 #include <chrono>
+#include <cstdio>
+#include <cstdlib>
 #include <filesystem>
 
-#include "src/decoders/novatel/api/commander.hpp"
-#include "src/hw_interface/stream_interface/api/outputfilestream.hpp"
+#include <decoders/novatel/api/commander.hpp>
+#include <hw_interface/stream_interface/api/outputfilestream.hpp>
```
jonathanmcdermid, line 33: why angled brackets?

regression/run_tests.sh
```diff
@@ -0,0 +1,36 @@
...
+    "$converter_components" "$script_dir/../database/messages_public.json" "$script_dir/BESTUTMBIN.GPS" $FORMAT > /dev/null
+    retval=$?
+    if [ $retval -ne 0 ]; then
+      echo "converter_components failed for $FORMAT failed with status $retval"
+    fi
+
+    if ! cmp -s "$script_dir/BESTUTMBIN.GPS.$FORMAT" "$script_dir/targets/BESTUTMBIN.GPS.$FORMAT"; then
+        failures+=("OEM-$FORMAT")
+    else
+      rm "$script_dir/BESTUTMBIN.GPS.$FORMAT"
```
jonathanmcdermid, line 22: is there a reason lines 20 and 22 have different indentations?

test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(test_package CXX)
+
+find_package(novatel_edie REQUIRED CONFIG)
+
+add_executable(CommandEncoding ${CMAKE_CURRENT_LIST_DIR}/../examples/novatel/command_encoding/command_encoding.cpp)
+target_link_libraries(CommandEncoding novatel_edie::novatel_edie)
+target_compile_features(CommandEncoding PRIVATE cxx_std_20)
```
jonathanmcdermid, line 8: does this need to be C++20?

vcpkg.json
```diff
@@ -0,0 +1,14 @@
+{
```
jonathanmcdermid, line 1: is vcpkg adding anything to this PR? If its not being utilized/tested in the pipeline and it isnt explained in the README, would it be better to leave it out for now?

.github/workflows/build.yml
valgur: Indeed, that was the intention. Makes it easier to test the changes to CI locally, before pushing.

CMakeLists.txt
```diff
@@ -1,18 +1,15 @@
...
-project(novatel-edie)
+include(cmake/third_party.cmake)
 
-enable_testing()
+project(novatel-edie VERSION 3.3.7 LANGUAGES CXX)
 
+option(BUILD_TESTS "Build tests" ON)
+option(BUILD_EXAMPLES "Build examples" ON)
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
+option(CMAKE_POSITION_INDEPENDENT_CODE "Set -fPIC" ON)
```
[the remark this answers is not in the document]
valgur: `CMAKE_POSITION_INDEPENDENT_CODE` is a CMake variable that enables the addition of `-fPIC` on non-Windows builds to the compiler commands. It's enabled more often than not in CMakeLists because static libraries built without `-fPIC` cannot be linked into shared libraries. The standard `fPIC` option in conanfiles also sets this value to true automatically as well. It also has some minor security benefits, I suppose, and can have a tiny performance overhead.

examples/novatel/command_encoding/command_encoding.cpp
valgur: Yeah, I like to think of example code files as standalone code projects that should work independently from the main code base, if possible. Double quotes would be ok as well, though.

regression/run_tests.sh
valgur: My bad. Thanks!

test_package/CMakeLists.txt
valgur: Forgot to update this one. Will fix this. Thanks!

vcpkg.json
valgur: Sure, I'll move it to a separate PR and add a CI pipeline for it.
I included it initially just in case you had not decided on which package manager you would like to incorporate into the repo, but that's no longer relevant.

jonathanmcdermid (approved):
looks good!

PR #24, opened 2024-05-03
Replace FileExists() with std::filesystem, avoid src/ prefix in includes

valgur:
A small and minor one.

The `src/` prefix should eventually be replaced with something package-specific, like `novatel_edie/`, `novatel/edie` or just `edie/`. This is just a small step in that direction.

Also fixed invalid number of bytes written in some of the C++ examples.


valgur:
Superseded by #42.

PR #25, opened 2024-05-03
Fix a memory leak in CircularBuffer

valgur:
- Add a destructor to properly clean up the buffer.
- `new` only returns a nullpointer as expected in the surrounding function if `std::nothrow` is used.


jonathanmcdermid:
Looks good!

PR #26, opened 2024-05-03
hw_interface: allow 'const char*' input, drop unused headers, throw for non-implemented methods

valgur:


jonathanmcdermid (approved):
looks good!

PR #30, opened 2024-05-03
bugfix: ASCII decoding can overflow input string

valgur:


src/decoders/common/src/message_decoder.cpp
```diff
@@ -398,8 +398,13 @@ STATUS MessageDecoderBase::DecodeAscii(const std::vector<BaseField*>& vMsgDefFie
     constexpr char acDelimiter3[4] = {cDelimiter1, cDelimiter2, cDelimiter3, '\0'};
     constexpr char acDelimiterResponse[2] = {cDelimiter2, '\0'};
 
+    char* pcBufEnd = *ppcLogBuf_ + strlen(*ppcLogBuf_);
+
     for (auto& field : vMsgDefFields_)
     {
+        if (*ppcLogBuf_ >= pcBufEnd) // We encountered the end of the buffer unexpectedly
+            return STATUS::MALFORMED_INPUT;
```
jonathanmcdermid, line 406:
```suggestion
        if (*ppcLogBuf_ >= pcBufEnd) { return STATUS::MALFORMED_INPUT; } // We encountered the end of the buffer unexpectedly
```

jonathanmcdermid (approved):
looks good!

PR #33, opened 2024-05-03
decoders: round instead of truncating when converting floats to int

valgur:
This caused millisecond values in the `ASCII_LOG_ROUNDTRIP_GLOALMANAC` test to diverge in the least significant digit after the round trip on some non-x86_64 architectures.


jonathanmcdermid (approved):
Looks good!

PR #35, opened 2024-05-03
Handle root logger not being initialized

valgur:
Call the default `InitLogger()` if `RegisterLogger()` is called without being initialized first.

Segfaults due to dereferencing a nullptr in the uninitialized root logger object otherwise.


jonathanmcdermid:
Looks good!

PR #38, opened 2024-05-04
Don't require git for version.h, autoformat, use project version from CMakeLists.txt

valgur:
The project should really be buildable from sources alone, when not cloned as a Git repo. Pretty much all package repositories rely on source tarballs to build the packages for distribution.


valgur:
@riley-kinahan What do you think about the changes made here? Seems somewhat relevant in preparation for the release in #117.

riley-kinahan:
@valgur 

In #117 I made version fully dependent on git to versioning, but I hadn't considered the value in making the project buildable from sources. I'll switch back to directly including the version in the main CMakeLists file and use a regex to manage it.

valgur:
Ok, nice. I'll update the PR to fix the version conflicts then asap as well.

PR #44, opened 2024-05-31
Fix fields with %e. Can be either float or double

rldoris:
Getting variant field access on fields with "%e" conversion string and type double.

Update message_decoder to check for field type first.


jonathanmcdermid (approved):
looks good!

PR #50, opened 2024-06-18
Fix undefined UINT_MAX

valgur:
Was getting undefined `UINT_MAX` errors on GCC.
Avoids `UINT_MAX` altogether as is not uniquely defined and depends on the size of `int`.

### DIFF
--- a/src/decoders/dynamic_library/src/novatel_file_parser.cpp
+++ b/src/decoders/dynamic_library/src/novatel_file_parser.cpp
@@ -24,9 +24,9 @@
 // ! \file novatel_file_parser.cpp
 // ===============================================================================
 
-#include "novatel_file_parser.hpp"
-
 #include <limits>
+
+#include "novatel_file_parser.hpp"
 
 using namespace novatel::edie;
 using namespace novatel::edie::oem;
--- a/src/decoders/dynamic_library/src/novatel_file_parser.cpp
+++ b/src/decoders/dynamic_library/src/novatel_file_parser.cpp
@@ -26,6 +26,8 @@
 
 #include "novatel_file_parser.hpp"
 
+#include <limits>
+
 using namespace novatel::edie;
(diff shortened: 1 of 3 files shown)


src/decoders/dynamic_library/src/novatel_parser.cpp
```diff
@@ -26,6 +26,8 @@
 
 #include "novatel_parser.hpp"
 
+#include <limits>
+
```
jonathanmcdermid, line 30:
```suggestion
#include <limits>

#include "novatel_parser.hpp"

```

src/decoders/dynamic_library/src/novatel_framer.cpp
```diff
@@ -26,6 +26,8 @@
 
 #include "novatel_framer.hpp"
 
+#include <limits>
+
```
jonathanmcdermid, line 30:
```suggestion
#include <limits>

#include "novatel_framer.hpp"

```

src/decoders/dynamic_library/src/novatel_file_parser.cpp
```diff
@@ -26,6 +26,8 @@
 
 #include "novatel_file_parser.hpp"
 
+#include <limits>
+
```
jonathanmcdermid, line 30:
```suggestion
#include <limits>

#include "novatel_file_parser.hpp"

```

jonathanmcdermid (approved):
looks good!

src/decoders/dynamic_library/src/novatel_parser.cpp
valgur, line 1: I would say the convention and best practice for headers corresponding to the specific source file is to place them at the top of the list of includes. This ensures that any include statements missing in the corresponding header file are not being accidentally provided by the include statements in the source file.

PR #51, opened 2024-06-18
Restore C++17 compatibility

valgur:
It's definitely up for debate whether C++20 features like ranges is something that you want or need, but as they are being used right now does not really add too much and can be avoided.
It's also worth keeping in mind that just an `#include <ranges>` (or an `#include <algorithm>` with C++20 enabled) by itself can significantly slow down the compilation of each translation unit due to the amount of templated code it includes: https://build-bench.com/b/hAQp8VHVgqIa9GC-suXYYxZmJLU


jonathanmcdermid (approved):
looks good!

PR #37, opened 2024-05-04
Normalize file endings in Git index

valgur:
The Git index in this repo currently contains a mix of source files with LF and CRLF line endings. This PR enables Git's autocrlf feature by adding `* text=auto` to `.gitattributes`. This forces all files in the Git index to use LF (which most of them already are do anyway). Checked are converted automatically to CRLF on Windows.

IDEs are usually smart enough to keep using the previous line endings in existing but can revert to the system default one when adding new files to the repo. It usually a good idea to prevent that by setting and forgetting the appropriate lines in .gitattributes.

To fix the Git index line endings in some other branch: `git add . --renormalize` and commit.
To fix the line endings in your working tree: `git rm --cached -r . && git reset --hard`

Here's a diff of `git ls-files --eol` before and after:
```diff
--- pre.list
+++ post.list
@@ -34,7 +34,7 @@
 i/lf    w/crlf  attr/text=auto        	examples/novatel/rxconfig_handler/rxconfig_handler.cpp
 i/lf    w/crlf  attr/text=auto        	lib/logger/example_logger_config.toml
 i/lf    w/crlf  attr/text=auto        	lib/logger/logger.hpp
-i/crlf  w/crlf  attr/text=auto        	lib/nlohmann/json.hpp
+i/lf    w/crlf  attr/text=auto        	lib/nlohmann/json.hpp
 i/lf    w/crlf  attr/text=auto        	lib/nlohmann/json_fwd.hpp
 i/lf    w/crlf  attr/text=auto        	lib/spdlog/async.h
 i/lf    w/crlf  attr/text=auto        	lib/spdlog/async_logger-inl.h
@@ -165,14 +165,14 @@
 i/lf    w/crlf  attr/text eol=crlf    	src/decoders/common/CMakeLists.txt
 i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/circular_buffer.hpp
 i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/common.hpp
-i/crlf  w/crlf  attr/text=auto        	src/decoders/common/api/crc32.hpp
+i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/crc32.hpp
 i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/encoder.hpp
 i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/framer.hpp
 i/lf    w/crlf  attr/text=auto        	src/decoders/common/api/json_reader.hpp
[...]

PR #53, opened 2024-06-25
Fix line endings for CMakeLists.txt and conanfile.txt

valgur:
Due to an oversight in #37, CMakeLists.txt and conanfile.txt are checked out as CRLF on non-Windows systems. This won't affect any branches or PRs, fortunately, since the files are already stored as LF in the Git index either way.


jonathanmcdermid (approved):
looks good!

PR #67, opened 2024-08-20
Delete Stream Interface

jonathanmcdermid:
Delete stream interface, use unique_ptr where possible and clean up some todos.


rldoris:
With the removal of the StreamInterface do we have any tests that check our unicode support? If not could you add them back in and just test opening the file with FileParser and getting a log out or something that reasonably shows that our interface supports unicode file paths?

rldoris (approved):

PR #73, opened 2024-09-19
Add support for statically linking dependencies

rldoris:
Add support for statically linking dependencies. Needed for Linux EDIE Python build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,14 +1,24 @@
 cmake_minimum_required(VERSION 3.15)
-include(cmake/third_party.cmake)
-
-project(novatel-edie VERSION 3.3.7 LANGUAGES CXX)
 
 option(BUILD_TESTS "Build tests" ON)
 option(BUILD_EXAMPLES "Build examples" ON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,14 +1,24 @@
...
+    unset(CONAN_INSTALL_ARGS CACHE)
+    set(CONAN_OPTIONS -o spdlog/*:shared=False -o fmt/*:shared=False)
+else()
+    # Set Conan dynamically link dependencies
+    unset(CONAN_INSTALL_ARGS CACHE)
+    set(CONAN_OPTIONS -o spdlog/*:shared=True -o fmt/*:shared=True)
+endif()
+
+include(cmake/third_party.cmake)
+project(novatel-edie VERSION 3.3.7 LANGUAGES CXX)
```
jonathanmcdermid, line 21: is this a better location than where it was before?

rldoris: Need to set the options before the conan import so Conan knows to statically link the dependencies. If the options are configured after the import then it's after the `conan install` command.

jonathanmcdermid (approved):
looks good!

PR #74, opened 2024-09-27
cppcoreguidelines

jonathanmcdermid:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

follow cppcoreguidelines more closely. Also made some changes to remove the '%' from conversion string hash calculation.

### DIFF
--- a/include/novatel_edie/decoders/common/json_reader.hpp
+++ b/include/novatel_edie/decoders/common/json_reader.hpp
@@ -48,25 +48,17 @@ using nlohmann::json;
 class JsonReaderFailure : public std::exception
 {
   private:
-    const char* func;
-    const char* file;
-    int32_t line;
-    std::filesystem::path clFilePath;
-    const char* failure;
-    char acWhatString[256]{};
+    std::string whatString;
 
   public:
-    JsonReaderFailure(const char* func_, const char* file_, const int32_t line_, std::filesystem::path jsonFile_, const char* failure_)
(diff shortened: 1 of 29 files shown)


include/novatel_edie/decoders/common/json_reader.hpp
```diff
@@ -258,52 +249,42 @@ struct BaseField
...
-
-            // If the value "10" or greater is found from the conversion string, two bytes would
-            // need to be consumed from the string to move past that value. Otherwise, only one byte
-            // is necessary to consume.
-            if (*sConvertString >= '0' && *sConvertString <= '9')
-            {
-                if (sscanf(sConvertString, "%d.", iSelectedPoint) != 1) { throw std::runtime_error("Failed to parse integer value"); }
-                sConvertString = sConvertString + (*iSelectedPoint > 9 ? 2 : 1); // Skip the numerals
-            }
+            conversionBeforePoint = std::stoi(sConvertString);
```
rldoris, line 261: This is a pretty significant change. You're sure its functionally equivalent?

jonathanmcdermid: It depends on what kind of conversion strings are valid. I think the way we were doing this before with the loop was nonsensical and bugprone. The way I wrote it, it assumes that a valid conversion string must come in the form of %(x)(.y)a with () denoting optional fields. The way we had it before, there were a lot of seemingly faulty conversion strings that would pass.

rldoris (approved):

PR #76, opened 2024-10-10
Add Python bindings

valgur:
Draft PR for the Python bindings. I would like to merge the changes from #67 as well before I'd consider it complete.

* Most of the bindings are straightforward mappings to the C++ API. The biggest difference is in the decoded messages, which expose the decoded data directly via properties or as a dictionary via a `.to_dict()` call.
* Ported over all of the test suite and examples.
* The bindings are as lightweight and intuitive as possible without sacrificing performance. Manual management of buffers is hidden, but copying of data is avoided wherever possible.
* The default JSON database is loaded automatically and shared as a singleton. The user can still load and specify their own DB, though.
* The project is installable from source directly via `pip install .`, with a compiler being the only requirement.
* For Python 3.12 and newer, the library is built against the stable CPython API, meaning the same wheels will be compatible with any future Python releases.
* The bindings are fast - the full test suite takes less than 50 ms to run on my machine.

Other notes:
* The biggest change to the core library is the replacement of raw pointers with `std::shared_pointer<...>` in the public APIs. It's almost impossible to correctly keep track of the lifetimes of the created objects across the Python interface otherwise.
* Includes a few fixes for clang-tidy-18 suggestions/warnings to the core library, such as replacing many nested ternary operators with switches.
* In addition to Windows, Linux and macOS, the CI also builds successfully for Linux ppc64le and Linux arm64, but these are disabled due to the very slow build with emulation.

Here's an example of the Python API in use:
```python
>>> import novatel_edie as ne

    raw_msg = bytes(...)

    header_decoder = ne.HeaderDecoder()
    message_decoder = ne.MessageDecoder()

    status, header, meta = header_decoder.decode(raw_msg)
    status.raise_on_error()

    body = raw_msg[meta.header_length:]
    status, message = message_decoder.decode(body, meta)
    status.raise_on_error()

>>> header
Header(message_id=1309, message_type=0, port_address=64, length=184, sequence=0, idle_time=131, time_status=200, week=2268, milliseconds=371455000.0, receiver_status=33554432, message_definition_crc=56297, receiver_sw_version=16407)

>>> meta
MetaData(message_name='GALINAVEPHEMERIS', format=novatel_edie.bindings.HEADERFORMAT.BINARY, measurement_source=novatel_edie.bindings.MEASUREMENT_SOURCE.PRIMARY, time_status=novatel_edie.bindings.TIME_STATUS.SATTIME, response=False, week=2268, milliseconds=371455000.0, binary_msg_length=184, length=0, header_length=28, message_id=1309, message_crc=56297)

>>> print(message.message_name, message.message_crc)
GALINAVEPHEMERIS 56297

>>> message.to_dict()
{'sat_id': 13,
 'e5b_health': 0,
 'e5b_dvs': 0,
 'GALINAVEPHEMERIS_reserved1': 0,
 'GALINAVEPHEMERIS_reserved2': 0,
 'e1b_health': 0,
 'e1b_dvs': 0,
 'GALINAVEPHEMERIS_reserved3': 0,
 'GALINAVEPHEMERIS_reserved4': 0,
 'io_dnav': 106,
 'sisa': 107,
 'inav_signal_type': 3,
 'toe': 370800,
 'toc': 370800,
 'm0': 0.7113800735269051,
 'delta_n': 2.377956194277532e-09,
 'ecc': 0.00017794407904148102,
 'root_a': 5440.609663009644,
 'i0': 0.9993512110806739,
 'i_dot': -4.064455015151444e-10,
 'omega0': 1.7141175566765459,
 'omega': 1.0178326326089204,
 'omega_dot': -5.393796101829271e-09,
 'cuc': -7.333233952522278e-06,
 'cus': 5.772337317466736e-06,
 'crc': 241.5,
 'crs': -157.03125,
 'cic': 6.332993507385254e-08,
 'cis': -1.3597309589385986e-07,
 'af0': -1.931004226207733e-05,
 'af1': -2.842170943040401e-14,
 'af2': 0.0,
 'e1_e5a_bgd': 5.3551048040390015e-09,
 'e1_e5b_bgd': 5.587935447692871e-09}

>>> message
<GALINAVEPHEMERIS sat_id=13, e5b_health=0, e5b_dvs=0, GALINAVEPHEMERIS_reserved1=0, GALINAVEPHEMERIS_reserved2=0,
e1b_health=0, e1b_dvs=0, GALINAVEPHEMERIS_reserved3=0, GALINAVEPHEMERIS_reserved4=0, io_dnav=106, sisa=107, inav_signal_type=3,
toe=370800, toc=370800, m0=0.7113800735269051, delta_n=2.377956194277532e-09, ecc=0.00017794407904148102,
root_a=5440.609663009644, i0=0.9993512110806739, i_dot=-4.064455015151444e-10, omega0=1.7141175566765459, omega=1.0178326326089204,
omega_dot=-5.393796101829271e-09, cuc=-7.333233952522278e-06, cus=5.772337317466736e-06, crc=241.5, crs=-157.03125, cic=6.332993507385254e-08,
cis=-1.3597309589385986e-07, af0=-1.931004226207733e-05, af1=-2.842170943040401e-14, af2=0.0, e1_e5a_bgd=5.3551048040390015e-09,
e1_e5b_bgd=5.587935447692871e-09> 

>>> message.e1_e5b_bgd
5.587935447692871e-09

>>> message._fields
{'sat_id': BaseField(name='sat_id', type=SIMPLE, data_type=ULONG, description='Sat Id', conversion='%u'),
 'e5b_health': BaseField(name='e5b_health', type=SIMPLE, data_type=UCHAR, description='E5b Health status', conversion='%UB'),
 'e5b_dvs': BaseField(name='e5b_dvs', type=SIMPLE, data_type=UCHAR, description='E5b Data validity status', conversion='%UB'),
 'GALINAVEPHEMERIS_reserved1': BaseField(name='GALINAVEPHEMERIS_reserved1', type=SIMPLE, data_type=UCHAR, description='Reserved', conversion='%UB'),
 'GALINAVEPHEMERIS_reserved2': BaseField(name='GALINAVEPHEMERIS_reserved2', type=SIMPLE, data_type=UCHAR, description='Reserved', conversion='%UB'),
 'e1b_health': BaseField(name='e1b_health', type=SIMPLE, data_type=UCHAR, description='E1b Health status', conversion='%UB'),
 'e1b_dvs': BaseField(name='e1b_dvs', type=SIMPLE, data_type=UCHAR, description='E1b Data validity status', conversion='%UB'),
 'GALINAVEPHEMERIS_reserved3': BaseField(name='GALINAVEPHEMERIS_reserved3', type=SIMPLE, data_type=UCHAR, description='Reserved', conversion='%UB'),
 'GALINAVEPHEMERIS_reserved4': BaseField(name='GALINAVEPHEMERIS_reserved4', type=SIMPLE, data_type=UCHAR, description='Reserved', conversion='%UB'),
 'io_dnav': BaseField(name='io_dnav', type=SIMPLE, data_type=USHORT, description='Issue of data ephemeris', conversion='%hu'),
 'sisa': BaseField(name='sisa', type=SIMPLE, data_type=UCHAR, description='Signal in space accuracy', conversion='%UB'),
 'inav_signal_type': BaseField(name='inav_signal_type', type=SIMPLE, data_type=UCHAR, description='Can be E1b, E5b, or E1b_E5b ', conversion='%UB'),
 'toe': BaseField(name='toe', type=SIMPLE, data_type=UINT, description='Ephemeris reference time (s)', conversion='%u'),
 'toc': BaseField(name='toc', type=SIMPLE, data_type=ULONG, description='Clock correction reference (s)', conversion='%u'),
 'm0': BaseField(name='m0', type=SIMPLE, data_type=DOUBLE, description='Mean anomaly at ref time (radians)', conversion='%0.9lk'),
 'delta_n': BaseField(name='delta_n', type=SIMPLE, data_type=DOUBLE, description='Mean motion difference (radians/s)', conversion='%0.4lk'),
 'ecc': BaseField(name='ecc', type=SIMPLE, data_type=DOUBLE, description='Eccentricity (unitless)', conversion='%0.9lk'),
 'root_a': BaseField(name='root_a', type=SIMPLE, data_type=DOUBLE, description='Semi-major axis (metres)', conversion='%0.9lk'),
 'i0': BaseField(name='i0', type=SIMPLE, data_type=DOUBLE, description='Inclination angle at ref time (radians)', conversion='%0.9lk'),
 'i_dot': BaseField(name='i_dot', type=SIMPLE, data_type=DOUBLE, description='Rate of inclination angle (radians/s)', conversion='%0.4lk'),
 'omega0': BaseField(name='omega0', type=SIMPLE, data_type=DOUBLE, description='Longitude of ascending node (radians)', conversion='%0.9lk'),
 'omega': BaseField(name='omega', type=SIMPLE, data_type=DOUBLE, description='Argument of perigee (radians)', conversion='%0.9lk'),
 'omega_dot': BaseField(name='omega_dot', type=SIMPLE, data_type=DOUBLE, description='Rate of right ascension (radians/s)', conversion='%0.7lk'),
 'cuc': BaseField(name='cuc', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the cosine harmonic correction term to the argument of latitude (radians)', conversion='%0.4lk'),
 'cus': BaseField(name='cus', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the sine harmonic correction term to the argument of latitude (radians)', conversion='%0.4lk'),
 'crc': BaseField(name='crc', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the cosine harmonic correction term to the orbit radius (m)', conversion='%0.4lk'),
 'crs': BaseField(name='crs', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the sine harmonic correction term to the orbit radius (m)', conversion='%0.4lk'),
 'cic': BaseField(name='cic', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the cosine harmonic correction term to the angle of inclination (radians)', conversion='%0.4lk'),
 'cis': BaseField(name='cis', type=SIMPLE, data_type=DOUBLE, description='Amplitude of the sine harmonic correction term to the angle of inclination (radians)', conversion='%0.4lk'),
 'af0': BaseField(name='af0', type=SIMPLE, data_type=DOUBLE, description='Clock correction constant term (s)', conversion='%0.9lk'),
 'af1': BaseField(name='af1', type=SIMPLE, data_type=DOUBLE, description='Clock correction 1st order term (s^-1)', conversion='%0.6lk'),
 'af2': BaseField(name='af2', type=SIMPLE, data_type=DOUBLE, description='Clock correction 2nd order term (s^-2)', conversion='%0.1lk'),
 'e1_e5a_bgd': BaseField(name='e1_e5a_bgd', type=SIMPLE, data_type=DOUBLE, description='E1E5a broadcast group delay', conversion='%0.3lk'),
 'e1_e5b_bgd': BaseField(name='e1_e5b_bgd', type=SIMPLE, data_type=DOUBLE, description='E1E5b broadcast group delay', conversion='%0.3lk')}
```


.github/workflows/build.yml
```diff
@@ -1,9 +1,16 @@
-name: Build
-
+name: C++
```
jonathanmcdermid, line 1: should the filename be updated to match this like the python.yml?

rldoris:
Hi @valgur,

I've been reviewing the changes, and there's quite a bit to take in, so I may have a few questions as I dig deeper—thanks in advance for your patience.

I've successfully compiled and run it locally and started speed tests. The performance looks excellent on the smaller datasets I had available, and I'm planning to test with larger, more diverse datasets for a broader assessment.

I noticed some changes in the Python interfaces, but a few features that were present before seem to be missing (no iteration over the parsers or framer, parsers not returning a PyDecodedMessage object, etc.). I wanted to check if that was intentional or if it's something that might need to be addressed. I assume this is an initial pass, and we can continue refining the interface after the PR to better suit our needs.

rldoris:
On a development branch of the python EDIE code I have ENUMs implemented at the message level so all ENUMS (e.g. `bestpos.position_type`) will return a ENUM like object that can be much easier to work with. Would it be possible to implement something like this with this using Nanobind? It was much easier to work with ENUMS. I'll share the code that enabled this.

common.py
```Python
# Code source: Patrick Roocks on Stack Overflow
# URL: https://stackoverflow.com/a/70877075

def typed_enumeration_type(tp):
    class EnumerationType(type(tp)):  # type: ignore
        def __new__(metacls, name, bases, dict):
            _members_ = dict.get("_members_")
            if _members_ is None:
                _members_ = {}
                for key, value in dict.items():
                    if not key.startswith("_"):
                        _members_[key] = value

                dict["_members_"] = _members_

            dict["_reverse_map_"] = {v: k for k, v in _members_.items()}
            cls = type(tp).__new__(metacls, name, bases, dict)
            return cls

        def __repr__(self):
            return "<Enumeration %s>" % self.__name__

    return EnumerationType


def typed_c_enumeration(tp=c_uint32):
    class CEnumeration(tp, metaclass=typed_enumeration_type(tp)):
        _members_ = {}

        def __repr__(self):
            value = self.value
            return f"<{self.__class__.__name__}.{self._reverse_map_.get(value, '(UNKNOWN)')}: {value}>"

        def __str__(self):
            value = self.value
            return f"{self.__class__.__name__}.{self._reverse_map_.get(value, '(UNKNOWN)')}"

        def __int__(self):
            return self.value

        def __eq__(self, other):
            if isinstance(other, int):
                return self.value == other

            if isinstance(other, str):
                return self._reverse_map_[self.value] == other

            return type(self) == type(other) and self.value == other.value

        @property
        def name(self):
            return str(self)

        def __iter__(self):
            self._iter_index = 0
            self._iter_keys = list(self._members_.keys())
            return self

        def __next__(self):
            if self._iter_index < len(self._iter_keys):
                key = self._iter_keys[self._iter_index]
                self._iter_index += 1
                return f'{self.__class__.__name__}.{key}'
            else:
                raise StopIteration

    return CEnumeration
```

jsonreader.py
```Python
    def _generate_enum_definitions(self, json_db: dict):
        for enum_def in json_db['enums']:
            enum_id = enum_def['_id']
            enum_name = enum_def['name']
            enum_values = {enum['name']: enum['value'] for enum in enum_def['enumerators']}
            self.enum_definitions[enum_id] = type(enum_name, (typed_c_enumeration(), ), enum_values)


    def _novtype_to_ctype(self, nov_field: dict, super_name: str) -> type:
        field_type = nov_field['type']
        if field_type in ['SIMPLE', 'STRING', 'RESPONSE_ID', 'RESPONSE_STR', 'FIXED_LENGTH_ARRAY',
                          'VARIABLE_LENGTH_ARRAY']:
            if nov_field['arrayLength']:
                # Ensure 4-byte alignment for fixed and variable-length arrays.
                padding = 0
                if (nov_field['arrayLength'] % 4) != 0:
                    padding = (4 - (nov_field['arrayLength'] % 4))
                if nov_field.get('conversionString', None) and '%s' in nov_field['conversionString']:
                    return typed_c_char_array(nov_field['arrayLength'] + padding)
                else:
                    return self.novtypes_to_ctypes_dict[nov_field['dataType']['name']] * (
                                nov_field['arrayLength'] + padding)
                # return self.novtypes_to_ctypes_dict[nov_field['dataType']['name']] * (nov_field['arrayLength'] + padding)
            else:
                return self.novtypes_to_ctypes_dict[nov_field['dataType']['name']]
        elif field_type == 'ENUM':
            try:
                return self.enum_definitions[nov_field['enumID']]
            except KeyError:
                return self.UNKNOWN_ENUM
        ...
```

User code now could look like this:
`if version.versions[0].component_type == 'GPSCARD'`

Would it be possible to implement something similar? I can share the code with you if you're interested in seeing it.

valgur:
@rldoris I agree, that would indeed make the Python bindings much more convenient to use.
I added some extra logic to map all of the JSON DB enums to `enum.IntEnum` types under `novatel_edie.message_enums.*` and added automatic conversion to the appropriate type in message decoding.

So a BESTPOS message looks something like this now:
```python
<BESTPOS solution_status=<SolStatus.SOL_COMPUTED: 0>, position_type=<SolType.SINGLE: 16>, latitude=51.15043711386, longitude=-114.03067767, orthometric_height=1097.2099, undulation=-17.0, datum_id=<Datum.WGS84: 61>, latitude_std_dev=0.9038000106811523, longitude_std_dev=0.8533999919891357, height_std_dev=1.7480000257492065, base_id='', diff_age=0.0, solution_age=0.0, num_svs=35, num_soln_svs=30, num_soln_L1_svs=30, num_soln_multi_svs=30, extended_solution_status2=0, ext_sol_stat=6, gal_and_bds_mask=57, gps_and_glo_mask=51>
```
and
```python
>>> list(ne.message_enums.SolStatus)
[<SolStatus.SOL_COMPUTED: 0>,
 <SolStatus.INSUFFICIENT_OBS: 1>,
 <SolStatus.NO_CONVERGENCE: 2>,
 <SolStatus.SINGULARITY: 3>,
 ...
 ```

The only downside is that a custom JSON DB can no longer really be used with the Python bindings since I resorted to using the JSON DB singleton as the source of the enum definitions to keep the bookkeeping manageable.

**Comment by rldoris** [comment]
I tested out the enums and they are working great. 

Would it be much effort to overload the equality operator to also compare against the enum name string? Totally not required, but it would make enum types easier to work with. Easy to implement in Python, but I attempted to implement in Nanobind and I failed to get anywhere. 

e.g. `bestpos.position_type == 'SINGLE'`

Is there someway to get around the lack of CustomDB support? Internally we have other databases for internal only logs and developers like to use custom JSON files when developing new logs. Developers could replace the JSON file in the resources directory, but that would be a bit of a pain for them. I'll take a look at the singleton tomorrow to try and get a better understanding of the issue.

**Comment by valgur** [comment]
@rldoris Regarding the custom DB support, I have three approaches we could consider:

- A quick fix would be to tweak the singleton to optionally use a DB from an environment variable instead loading the default DB. This might be sufficient for development purposes, but is probably not ideal.

- Add a `novatel_edie.load_database(...)` that overrides the default one and replaces the contents of `novatel_edie.message_enums.*`.

- Probably the best choice would be to extend the `JsonDb` bindings, so it would have an `.enums` property with the same content as `novatel_edie.message_enums` and the DB object could then be passed to any Decoder/Encoder object constructor. The `PyDecodedMessage` would then likely need to keep track of the associated DB, though, which is a bit tricky and could lead to reference cycles.

Regarding the support for string comparison, extending the IntEnum base class would actually be a bit tricky since these enums are being constructed on the C++ side. I would say these two alternatives are good enough, though?

```python
bestpos.position_type.name == 'SINGLE'
bestpos.position_type == ne.message_enums.SolType.SINGLE
```

Also... I now noticed that `from novatel_edie.message_enums import SolType` does not work due to nanobind limitations (https://github.com/pybind/pybind11/issues/2639). :slightly_frowning_face: I could maybe rename it the submodule to just `enums` so that the more concise `ne.enums.SolType` would work. I did not go with that to avoid confusion with all of the other enums. Or I could simply dump all of the dynamic enum definitions into the root module, which seemed a bit messy to me (i.e. it would make autocomplete harder to use due to too many symbols being listed in the root module).

rldoris:
I agree that the 3rd option seems like the best choice. I like the idea of continuing to give the option of passing the JsonDB into the constructor as I can think of a few use cases where that might be useful to have different databases. Embedding the ENUM definitions with the JsonDB object feels reasonable. I don't know if I've seen a python package do that before, but I can't think of a more elegant method to achieve behavior. 

On a side note: one thing I would like to try in the future is to separate the JSON file into individual JSON files for each release of our FW. Currently I'm generating JsonDB for each FW release and then smashing them together to create a single Json file. While this mostly works I it feels a bit crude and there are some potential issues around message/enum definitions changing between different releases (among other issues). If we had different JsonDB files for each release then developers could create multiple JsonDB objects (for each FW) and then decode the messages specifically for that release. This would also remove some of the cruft from the JSON like `latestMsgDefCrc` and the array of `fields`.

The 2 alternatives are more than good enough. I think most people will default to using the `.name  ==` comparison so I was thinking about trying to optimize that as much as possible, but it isn't worth the effort of trying to over optimize it.

valgur:
A summary of the latest changes:
1. I refactored JsonReader into a MessageDatabase and JsonDbReader classes. Most of the code in NovAtel EDIE cares only about the database aspect of JsonReader and not about the storage/serialization details of it, so it made sense to me to keep them separate. The code between was already effectively decoupled, I just needed to move things around a bit. This change was motivated by making the MessageDatabase a base class that could be extended slightly into a PyMessageDatabase class for the Python enumerations handling. The JSON parsing aspect was superfluous there. The decoupling might also be useful for swapping the JSON serialization format out for something more optimized, e.g. some alternative format that could be generated from the JSON DB as part of the build. Let me know if you have any doubts about this refactoring.
2. The Python bindings now only output and take as input PyMessageDatabase objects, which hold the additional Python IntEnum mappings needed for decoded messages. The decoded messages in turn keep track of the MessageDatabase they were generated from to apply that IntEnum info as necessary. The rest of the C++ library does not need to know anything about the derived PyMessageDatabase class and only sees the MessageDatabase base class thanks to C++ polymorphism magic.
3. I replaced all `const char*`, `std::string` and `std::u32string` path input arguments in overloaded methods with just `std::filesystem::path`. The path type natively supports all Unicode C++ string types, such as `std::u32string` or `std::u8string`.
4. Renamed `novatel_edie.message_enums` to `novatel_edie.enums` due to `from novatel_edie.message_enums import ...` not being supported. This keeps the enum references in Python code a bit shorter.
5. I added an `.enums` property to the MessageDatabase Python bindings to match the `novatel_edie.enums.*`, which is populated by the enums from the default DB. I was not able to make them accessible as attributes/properties, though, so they can only be used as `db.enums["Datums"].WGS84` instead of `db.enums.Datums.WGS84`, for example.

src/decoders/oem/test/oem_test.cpp
```diff
@@ -1027,9 +1027,9 @@ class DecodeEncodeTest : public ::testing::Test
         int32_t iRetCode = DecodeEncode(eFormat_, pstExpectedMessageData_->pucMessage, pucEncodeBuffer, MAX_ASCII_MESSAGE_LENGTH, stMetaData, stMessageData);
 
         return iRetCode != SUCCESS                                                                                   ? iRetCode
-               : eFormat_ != ENCODE_FORMAT::JSON && stMetaData.uiHeaderLength != stMessageData.uiMessageHeaderLength ? HEADER_LENGTH_ERROR
-               : pstExpectedMessageData_->uiMessageLength != stMessageData.uiMessageLength                           ? LENGTH_ERROR
-               : stMessageData != *pstExpectedMessageData_                                                           ? MESSAGE_DATA_COMPARISON_ERROR
+               : eFormat_ != ENCODE_FORMAT::JSON && stMetaData.uiHeaderLength != stMessageData.uiMessageHeaderLength ? HEADER_LENGTH_ERROR // NOLINT
+               : pstExpectedMessageData_->uiMessageLength != stMessageData.uiMessageLength                           ? LENGTH_ERROR // NOLINT
+               : stMessageData != *pstExpectedMessageData_                                                           ? MESSAGE_DATA_COMPARISON_ERROR // NOLINT
```
jonathanmcdermid, line 1032: why NOLINT?

src/decoders/common/src/message_decoder.cpp
```diff
@@ -731,20 +736,28 @@ MessageDecoderBase::Decode(const unsigned char* pucMessage_, std::vector<FieldCo
...
+    case HEADER_FORMAT::ABB_ASCII: [[fallthrough]];
+    case HEADER_FORMAT::SHORT_ABB_ASCII: //
+        return DecodeAscii<true>(pvCurrentMsgFields, reinterpret_cast<const char**>(&pucTempInData), stInterMessage_);
+    case HEADER_FORMAT::BINARY: [[fallthrough]];
+    case HEADER_FORMAT::SHORT_BINARY: //
+        return DecodeBinary(pvCurrentMsgFields, &pucTempInData, stInterMessage_, stMetaData_.uiBinaryMsgLength);
+    case HEADER_FORMAT::JSON: //
+        return DecodeJson(pvCurrentMsgFields, json::parse(pucTempInData)["body"], stInterMessage_);
+    default: //
+        return STATUS::UNKNOWN;
```
jonathanmcdermid, line 761:
```suggestion
    case HEADER_FORMAT::ASCII: [[fallthrough]];
    case HEADER_FORMAT::SHORT_ASCII: return DecodeAscii<false>(pvCurrentMsgFields, reinterpret_cast<const char**>(&pucTempInData), stInterMessage_);
    case HEADER_FORMAT::ABB_ASCII: [[fallthrough]];
    case HEADER_FORMAT::SHORT_ABB_ASCII: return DecodeAscii<true>(pvCurrentMsgFields, reinterpret_cast<const char**>(&pucTempInData), stInterMessage_);
    case HEADER_FORMAT::BINARY: [[fallthrough]];
    case HEADER_FORMAT::SHORT_BINARY: return DecodeBinary(pvCurrentMsgFields, &pucTempInData, stInterMessage_, stMetaData_.uiBinaryMsgLength);
    case HEADER_FORMAT::JSON: return DecodeJson(pvCurrentMsgFields, json::parse(pucTempInData)["body"], stInterMessage_);
    default: return STATUS::UNKNOWN;
```

valgur: I'd keep it as is. The empty comments are there to ensure the autoformatter formats the non-fallthrough cases consistently for readability.

src/decoders/oem/test/oem_test.cpp
valgur: It's for `*-avoid-nested-conditional-operator` checks on newer clang-tidy versions. Should I disable it in `.clang-tidy` instead? I already converted all nested ternary operators to switches where possible in this PR.

jonathanmcdermid: ahh I see. I think the change to using switches is for the best. I dont see .clang-tidy showing up in the diff though, so why is this NOLINT necessary now when it wasnt before?

valgur: I'm assuming this is due to using clang-tidy 19 on my dev machine while the version in CI is 17. So the NOLINTs are currently not really needed as far as the CI is concerned.

rldoris:
@valgur Sorry for taking so long. 
I was adding comments and hadn't realized that they weren't publishing in GitHub. I'll remove any old or out of date comments that were just posed and keep working on the review.

python/bindings/novatel/parser.cpp
```diff
@@ -0,0 +1,54 @@
...
+        .def("enable_framer_decoder_logging", &oem::Parser::EnableFramerDecoderLogging, "level"_a = spdlog::level::debug, "filename"_a = "edie.log")
+        .def_prop_rw("ignore_abbreviated_ascii_responses", &oem::Parser::GetIgnoreAbbreviatedAsciiResponses,
+                     &oem::Parser::SetIgnoreAbbreviatedAsciiResponses)
+        .def_prop_rw("decompress_range_cmp", &oem::Parser::GetDecompressRangeCmp, &oem::Parser::SetDecompressRangeCmp)
+        .def_prop_rw("return_unknown_bytes", &oem::Parser::GetReturnUnknownBytes, &oem::Parser::SetReturnUnknownBytes)
+        .def_prop_rw("encode_format", &oem::Parser::GetEncodeFormat, &oem::Parser::SetEncodeFormat)
+        .def_prop_rw("filter", &oem::Parser::GetFilter, &oem::Parser::SetFilter)
+        .def("write",
+             [](oem::Parser& self, const nb::bytes& data) { return self.Write(reinterpret_cast<const uint8_t*>(data.c_str()), data.size()); })
+        .def("read",
```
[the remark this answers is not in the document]
rldoris: Thinking more about this comment I wonder if, in the future, it would be better to separate conversion (binary -> ascii) from programmatic access. The current API is a bit confusing where if you don't convert to flattened_binary the message variable is a None. Future thing to think about.

python/bindings/novatel/py_decoded_message.hpp
```diff
@@ -0,0 +1,45 @@
...
+    size_t len() const;
+    std::string repr() const;
+
+    std::vector<FieldContainer> fields;
+
+    // MetaDataStruct with all fields that are no longer relevant after decoding dropped.
+    uint16_t message_id;
+    uint32_t message_crc;
+    std::string message_name;
+    PyGpsTime time;
```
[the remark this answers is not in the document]
valgur: Most of the details in the MetaData object, besides the ones included here, are of no use to the user after the message has been parsed, really, and passing around a pair of objects in the consumer code seems unnecessarily clumsy to me. I originally only wanted to keep track of the message ID / name for a clearer repr(), but it made sense to copy all relevant fields instead.

I would suggest only returning the PyDecodedMessage from Parser and FileParser, which provide a more abstract interface to the decoding process anyway, and would only return a pair of PyDecodedMessage and MetaData from the more advanced Decoder class.

What do you think?

Also, all of the message names should easily fit within the small string optimization limit of 23 characters, so they are stored in-place rather than on the heap for `message_name`, which is nice. I have not done any serious benchmarking, but I would expect the performance impact of copying over these fields to be negligible and the UI/UX improvement worth the cost.

python/bindings/py_decoded_message.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+    std::vector<FieldContainer> fields;
+
+    // MetaDataStruct with all fields that are no longer relevant after decoding dropped.
+    uint16_t message_id;
+    uint32_t message_crc;
+    std::string message_name;
+    PyGpsTime time;
+    MEASUREMENT_SOURCE measurement_source{MEASUREMENT_SOURCE::PRIMARY};
+    CONSTELLATION constellation{CONSTELLATION::UNKNOWN};
```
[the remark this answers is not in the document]
valgur: Yeah, I only included these for completeness. See the other comment. The measurement source is quite relevant in a dual-antenna setup, though, which seems to be common for receivers mounted on vehicles.

python/bindings/novatel/parser.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "json_db_singleton.hpp"
+#include "py_message_data.hpp"
+
+namespace nb = nanobind;
+using namespace nb::literals;
+using namespace novatel::edie;
+
+void init_novatel_parser(nb::module_& m)
+{
+    nb::class_<oem::Parser>(m, "Parser")
```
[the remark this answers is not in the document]
valgur: That's a very good suggestion. I'll look into it.

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -41,19 +45,23 @@ struct FieldContainer;
 
 #define NOVATEL_TYPES bool, int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t, float, double, std::string
 #define CONTAINER_TYPES NOVATEL_TYPES, std::vector<FieldContainer>
+using FieldValueVariant = std::variant<CONTAINER_TYPES>;
```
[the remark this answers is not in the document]
jonathanmcdermid: I recall we used to have an alias for a vector of fieldcontainers and I removed it because I thought it was a simple structure and more didactic to just write it as the underlying type.

python/bindings/novatel/py_decoded_message.hpp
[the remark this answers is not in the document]
rldoris: My initial thoughts:

I agree that the metadata is mostly useless after decoding and filtering except for the message name. Embedding that in the header or or body would be ideal. I checked internally and I can see message names up to a maximum of 31 characters so if we could make it that length (or a few extra bytes for safety) is very reasonable idea. I do think that the header should be passed around with the message data as there are a number of fields that are useful (id, port, idle, time status, time, and receiver status are all commonly used either in development or testing) and I think embedding all of them into the message would be a bit too much. Does it make sense to pass around a message object with a Header and Body?

For Parser and FileParser I think we still need the ability to return the header also as it provides critical fields that we need for our testing. I've tried pushing the more advanced framer->headerdecoder->filter->message decoder internally and it didn't get much traction. People much preferred the simpler interface of the Parser and FileParser. I think this behavior would extend to people/companies using this package to so I think providing the Header back, in addition to the body, would be required. We can discuss this further if you disagree.

I agree that the impact of storing the name with the message data (I prefer the header!) would be worth the cost. If we drop the metadata at some point in the process then the Python code gets much worse.

`header.name == 'bestpos'` looks much better than `header.id == 42'`

rldoris:
I'm seeing lots of exceptions when inspecting enums

```
db.enums['Datum'].ADIM
<Datum.ADIM: 65>
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_comm.py", line 1899, in do_it
    var_obj = pydevd_vars.getVariable(self.thread_id, self.frame_id, scope, attrs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_vars.py", line 242, in getVariable
    var = resolver.resolve(var, k)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.1\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_resolver.py", line 224, in resolve
    raise UnableToResolveVariableException()
_pydevd_bundle.pydevd_resolver.UnableToResolveVariableException
```

rldoris:
I've also encountered another build issue that I'm struggling to solve: 
When I build the package using the instructions in pyproject
```
$ pip install --upgrade nanobind scikit-build-core[pyproject]
$ pip install --no-build-isolation -ve .
```
I get a DLL error when importing the package. I tracked it down to `edie_oem_decoder.dll` missing from the directory with `bindings.cp311-win_amd64.pyd` in it.

When I build the project with Visual Studio the dependency is in the `bindings.cp311-win_amd64.pyd` file.


`pip install`
```
C:\Program Files\Microsoft Visual Studio\2022\Professional>dumpbin /dependents C:\Users\rdoris\source\repos\github\novatel_edie\venv\Lib\site-packages\novatel_edie\bindings.cp311-win_amd64.pyd
Microsoft (R) COFF/PE Dumper Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file C:\Users\rdoris\source\repos\github\novatel_edie\venv\Lib\site-packages\novatel_edie\bindings.cp311-win_amd64.pyd

File Type: DLL

  Image has the following dependencies:

    edie_oem_decoder.dll
    python311.dll
    KERNEL32.dll
    MSVCP140.dll
    VCRUNTIME140_1.dll
    VCRUNTIME140.dll
    api-ms-win-crt-heap-l1-1-0.dll
    api-ms-win-crt-runtime-l1-1-0.dll
    api-ms-win-crt-string-l1-1-0.dll
    api-ms-win-crt-stdio-l1-1-0.dll
    api-ms-win-crt-convert-l1-1-0.dll
    api-ms-win-crt-math-l1-1-0.dll
    api-ms-win-crt-filesystem-l1-1-0.dll
    api-ms-win-crt-locale-l1-1-0.dll
    api-ms-win-crt-time-l1-1-0.dll

  Summary

        B000 .data
        9000 .pdata
       3B000 .rdata
        2000 .reloc
        1000 .rsrc
       C2000 .text

```

`msvc`
```
C:\Program Files\Microsoft Visual Studio\2022\Professional>dumpbin /dependents C:\Users\rdoris\source\repos\github\nanob
ind_branch\out\build\x64-Debug\bin\windows-AMD64-msvc-Debug\bindings.cp311-win_amd64.pyd
Microsoft (R) COFF/PE Dumper Version 14.39.33523.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file C:\Users\rdoris\source\repos\github\nanobind_branch\out\build\x64-Debug\bin\windows-AMD64-msvc-Debug\bindings.cp311-win_amd64.pyd

File Type: DLL

  Image has the following dependencies:

    python311.dll
    KERNEL32.dll
    MSVCP140D.dll
    VCRUNTIME140D.dll
    VCRUNTIME140_1D.dll
    ucrtbased.dll

  Summary

        1000 .00cfg
       12000 .data
        7000 .idata
       58000 .pdata
      15E000 .rdata
        A000 .reloc
        1000 .rsrc
      3FB000 .text
        1000 .tls
```

Any thoughts on why my pip installed pyd file isn't building the dependencies into the binary (or why it isn't copying it during the install process)?

rldoris (approved):

python/bindings/novatel/parser.cpp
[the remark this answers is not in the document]
valgur: Done.

valgur:
I cannot reproduce the PyCharm resolver error on my end.
I suspect it's a PyCharm bug. The IDE version looks quite old, though (2020.3.1 vs the latest 2024.3), so an upgrade might fix it.

python/bindings/message_database.cpp
```diff
@@ -0,0 +1,224 @@
...
+        .def(nb::init())
+        .def(nb::init<std::string, uint32_t, std::string>(), "name"_a, "value"_a, "description"_a = "")
+        .def_rw("value", &EnumDataType::value)
+        .def_rw("name", &EnumDataType::name)
+        .def_rw("description", &EnumDataType::description)
+        .def("__repr__", [](const EnumDataType& enum_data_type) {
+            if (enum_data_type.description.empty())
+                return nb::str("EnumDataType(name={!r}, value={!r})").format(enum_data_type.name, enum_data_type.value);
+            return nb::str("EnumDataType(name={!r}, value={!r}, description={!r})")
+                .format(enum_data_type.name, enum_data_type.value, enum_data_type.description);
```
jonathanmcdermid, line 62:
```suggestion
            return enum_data_type.description.empty()
                ? nb::str("EnumDataType(name={!r}, value={!r})").format(enum_data_type.name, enum_data_type.value)
                : nb::str("EnumDataType(name={!r}, value={!r}, description={!r})").format(enum_data_type.name, enum_data_type.value, enum_data_type.description);
```

src/decoders/oem/src/encoder.cpp
```diff
@@ -118,44 +123,44 @@ void Encoder::InitFieldMaps()
...
+                                                 [[maybe_unused]] MessageDatabase& pclMsgDb_) {
         const auto uiTempId = std::get<uint32_t>(fc_.fieldValue);
         const uint16_t usSv = uiTempId & 0x0000FFFF;
         const int16_t sGloChan = (uiTempId & 0xFFFF0000) >> 16;
-        return sGloChan < 0    ? PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u%d")", usSv, sGloChan)
-               : sGloChan != 0 ? PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u+%d")", usSv, sGloChan)
-                               : PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u")", usSv);
+        if (sGloChan < 0) { return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u%d")", usSv, sGloChan); }
+        if (sGloChan != 0) { return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u+%d")", usSv, sGloChan); }
+        return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, R"("%u")", usSv);
```
jonathanmcdermid, line 142: what's the advantage with this change?

valgur: I think the ternary operators were failing clang-tidy validation on CI, which is why I replaced them. Can probably be reverted.

rldoris:
@valgur I'm pretty happy with where things are. The various issues have gone away when my computer died and I needed to rebuild everything from scratch so it's most likely a me issue somewhere. 

I've been trying to review both the binding and the Python interface changes and it's proven to be a lot to tackle with everything else I have going on. Would you mind if I approved this PR and then opened up an issue where we could discuss and come up with an agreement on the Python interface? There are a few items (e.g. the Parser and Fileparser not returning messages) that we use currently and would like to bring back in some form (different is fine, but internally people use the Fileparser to get fields from logs so I think that functionality should remain).

valgur:
@rldoris Sounds good. I'm definitely open to discussing and improving the interface design of the bindings API. They are currently mostly a direct mapping to the C++ one, which does not necessesarily make sense in all places. Merging it would also make my life a bit easier by not having  to continually fix merge conflicts from unrelated changes or having to apply CI fixes, etc, which should really go in a separate PR.

rldoris (approved):
The new nanobind plumbing looks great. Still concerns about the interface, but let's iron them out when we get the plumbing in place.

jonathanmcdermid:
@valgur how do you want to handle merging this? there are a ton of commits so the rebase looks rather painful. are you going to squash the commits?

valgur:
I would use a regular merge? The individual commits are meaningful and possibly useful for git blame, so I would avoid squashing. Rebasing is an option too, since I have done it several times already anyway and with even bigger changes, I think.

PR #77, opened 2024-10-23
Fix decoding of %c (used in BESTUTM)

valgur:
The `%c` conversion string is always being incorrectly decoded as 0, which gets converted to '\0' when re-encoding as ASCII or JSON.

I already applied the fix in #76.

The decoding and encoding would benefit from better validation overall. Both the decoder and encoder should check that the letter is printable or even strictly alphanumeric in this case, and many of the ASCII decoders use `strtoul` or similar, which silently returns a zero on decoding errors, etc.


rldoris:
@valgur Can we close this now that #76 has been merged?

valgur:
Right, we can.

PR #82, opened 2024-11-13
Fix Conan version conflicts

valgur:
Fixes the build and ensures that the build is not affected by recipe changes on ConanCenter in the future.

https://docs.conan.io/2/tutorial/versioning/lockfiles.html

The lockfile can be updated with either
```
conan lock create . -r conancenter
```
or
```
conan install . -r conancenter --lockfile-out=conan.lock --lockfile-clean
```

---
Also makes use of the `auto_shared_fpic` feature to simplify the recipe a bit (https://docs.conan.io/2/reference/conanfile/methods/configure.html#auto-shared-fpic)
and drops a `compiler.cppstd` that is no longer necessary in Conan v2.


conanfile.py
```diff
@@ -26,33 +26,29 @@ class NovatelEdieConan(ConanFile):
...
-            self.options.rm_safe("fPIC")
             self.options["spdlog"].shared = True
 
     def layout(self):
         cmake_layout(self)
 
     def requirements(self):
         self.requires("nlohmann_json/[>=3.11 <3.12]", transitive_headers=True)
-        self.requires("spdlog/[>=1.13 <2]", transitive_headers=True, transitive_libs=True)
+        self.requires("spdlog/[>=1.13 <2]", transitive_headers=True, transitive_libs=True, force=True)
```
jonathanmcdermid, line 46: why not just remove this line and allow spdlog-setup to install the version of spdlog that it requires? (1.14.1)

jonathanmcdermid:
why do we have ranges for the versions of the dependencies in the conanfile.py, but lock ourselves to a specific version in the lockfile? Can this not be accomplished by using discrete versions in the conanfile requirements?

conanfile.py
valgur: It's because conflict resolution for versions is almost non-existent for Conan. The version range effectively selects 1.15 while `spdlog_setup` requires 1.14, and it gives up at that point, declaring an unresolvable version conflict.
So the options are:
1. Use a version specific version that has no conflicts. The versions won't be latest then and a lockfile or a force=True is required to ensure stability against version bumps in upstream recipes.
2. Use a version range with force=True.

valgur:
> why do we have ranges for the versions of the dependencies in the conanfile.py, but lock ourselves to a specific version in the lockfile? Can this not be accomplished by using discrete versions in the conanfile requirements?

It's to ensure that the builds are reproducible and that there are no unexpected build failures for consumers.As I mentioned in the other comment, Conan is very prone to version conflicts, unfortunately.

The lockfile is a bit redundant with the force=True, though. I'll keep it simple and drop it.

PR #84, opened 2024-12-20
Framer Manager

zainarnoldhexagon:


src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+using namespace novatel::edie;
+
+FramerManager::FramerManager() : pclMyLogger(Logger::RegisterLogger("FramerManager")), pclMyCircularDataBuffer(std::make_shared<CircularBuffer>())
+{
+    pclMyCircularDataBuffer->Clear();
+    pclMyLogger->debug("Framer Manager initialized");
+}
+
+std::shared_ptr<CircularBuffer> FramerManager::GetCircularBuffer() const { return pclMyCircularDataBuffer; }
```
jonathanmcdermid, line 37: basic function definitions that are unlikely to ever change should be in the header file

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+void FramerManager::RegisterFramer(const FRAMER_ID framerId_, std::unique_ptr<FramerBase> framer_, std::unique_ptr<MetaDataBase> metadata_)
+{
+    framerRegistry.emplace_back(FramerElement(framerId_, std::move(framer_), std::move(metadata_), 0));
```
jonathanmcdermid, line 41: get rid of the temporary object here, just provide the arguments to emplace_back

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        HandleUnknownBytes(pucFrameBuffer_, framerRegistry.front().framer->uiMyFrameBufferOffset);
+        ResetAllFramerStates();
+        return STATUS::UNKNOWN;
+    }
+
+    // Once active framer is identified, exclusively use it to frame. Reset active Framer at success
+    for (auto& framer_element : framerRegistry)
+    {
+        if (framer_element.framerId == eActiveFramerId_)
```
jonathanmcdermid, line 191: if the framerId has a hit here, shouldn't the loop end?

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+            {
+                HandleUnknownBytes(pucFrameBuffer_, framer_element.framer->uiMyFrameBufferOffset);
+                ResetAllFramerStates();
+                return framer_element.framer->eMyCurrentFramerStatus;
+            }
+        }
+    }
+
+    //// A Framer Successfully Framed a log
+    auto& it_success = std::find_if(framerRegistry.begin(), framerRegistry.end(),
```
jonathanmcdermid, line 207: since we already found the active framer in the previous loop, cant we just use a reference to it instead of looking through the registry to find a success?

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,293 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+}
+
+// -------------------------------------------------------------------------------------------------------
+STATUS Framer::FindNextSyncByte(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSize_)
 {
     if (pucFrameBuffer_ == nullptr) { return STATUS::NULL_PROVIDED; }
 
-    // Loop buffer to complete NovAtel message
-    while (eMyFrameState != NovAtelFrameState::COMPLETE_MESSAGE)
+    enum class HeaderFormat
```
jonathanmcdermid, line 115: delete this?

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,293 @@ bool Framer::IsAbbrevAsciiResponse() const
...
     {
-        stMetaData_.bResponse = false;
+        UNKNOWN,
+        ABB_ASCII
+    };
+    NovAtelFrameState localFrameState = NovAtelFrameState::WAITING_FOR_SYNC;
 
+    while (localFrameState != NovAtelFrameState::COMPLETE_MESSAGE)
+    {
         // Read data from circular buffer until we reach the end or we didn't find a complete frame in current data buffer
```
jonathanmcdermid, line 124: this wording is confusing

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+    {
+        framerRegistry.front().framer->eMyCurrentFramerStatus = STATUS::INCOMPLETE_MORE_DATA;
+    }
+
+    if (eActiveFramerId_ == FRAMER_ID::UNKNOWN)
+    {
+        ResetAllMetaDataStates();
+        for (auto& [framerId, framer, metadata, offset] : framerRegistry)
+        {
+            framer->eMyCurrentFramerStatus = framer->FindNextSyncByte(pucFrameBuffer_, uiFrameBufferSize_);
```
jonathanmcdermid, line 163: make findNextSyncByte void since the return value is only being used to modify a member of the class

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,293 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+{
+    uiMyByteCount = 0;
+    uiMyExpectedMessageLength = 0;
+    uiMyExpectedPayloadLength = 0;
+    uiMyFrameBufferOffset = 0;
+    ResetState();
+}
+
+// -------------------------------------------------------------------------------------------------------
+STATUS Framer::FindNextSyncByte(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSize_)
```
jonathanmcdermid, line 111: why does this function take pucFrameBuffer as an argument?

src/common/src/framer_manager.cpp
zainarnoldhexagon: moved to header!

src/common/src/framer_manager.cpp
zainarnoldhexagon: yes, thanks for catching!

src/decoders/oem/src/framer.cpp
zainarnoldhexagon: Yeah, was an artifact of a previous paradigm

src/decoders/oem/src/framer.cpp
zainarnoldhexagon: The comment doesn't actually say what is happening. A decision within the if statement will choose if we have an incomplete log. I'll adjust here and on the NMEA framer

src/decoders/oem/src/framer.cpp
zainarnoldhexagon: It doesn't need to, I don't think. You're right!

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    // Once active framer is identified, exclusively use it to frame. Reset active Framer at success
+    for (auto& framer_element : framerRegistry)
+    {
+        if (framer_element.framerId == eActiveFramerId_)
+        {
+            ResetInactiveFramerStates(framer_element.framerId);
+            framer_element.framer->eMyCurrentFramerStatus =
+                framer_element.framer->GetFrame(pucFrameBuffer_, uiFrameBufferSize_, *framer_element.metadata);
```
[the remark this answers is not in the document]
jonathanmcdermid: Yeah I get that the circular buffer and the framer buffer are two separate things. My question is more about what can happen when we call GetFrame. Each framer is capable of changing the shared circular buffer (and in some cases, they do so), which seems like it could cause some unusual behaviour. I imagine this means that you can get different results depending on the order that you choose to call GetFrame for the different framers in the registry.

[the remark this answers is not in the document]
zainarnoldhexagon: I'll do a little investigation in this. You're right, there are a few places where this exists. I'll have to see why, because, you're right, the intent is to have the framer manager be the only modifier of the circular buffer

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,234 @@
...
+uint32_t FramerManager::Write(const unsigned char* pucDataBuffer_, uint32_t uiDataBytes_)
+{
+    return pclMyCircularDataBuffer->Append(pucDataBuffer_, uiDataBytes_);
+}
+
+[[nodiscard]] std::shared_ptr<spdlog::logger> FramerManager::GetLogger() const { return pclMyLogger; }
+
+void FramerManager::SetLoggerLevel(const spdlog::level::level_enum eLevel_) const { pclMyLogger->set_level(eLevel_); }
+
+void FramerManager::ShutdownLogger() { ShutdownLogger(); }
```
jonathanmcdermid, line 51: I would say all of these, and potentially registerframer too, are functions that really wont ever change in behaviour and belong in the header file

src/common/src/framer_manager.cpp
zainarnoldhexagon: Having the active framer doesn't mean that a frame will be successfully found. The active framer can return statuses indicating other outcomes, like BUFFER_FULL, UNKNOWN, NULL_PROVIDED, INCOMPLETE

src/common/src/framer_manager.cpp
zainarnoldhexagon: Fixed!

src/common/src/framer_manager.cpp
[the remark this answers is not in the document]
zainarnoldhexagon: I think the solution I'm about to upload should fix this!

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+// #include "src/decoders/nmea/api/framer.hpp"
+//
+// #include "novatel-edie/src/decoders/novatel/api/common.hpp"
+// #include "novatel-edie/src/decoders/novatel/api/framer.hpp"
+//
+// #include "src/decoders/pimtp/api/common.hpp"
+// #include "src/decoders/pimtp/api/framer.hpp"
+//
+// #include "src/decoders/waas/api/common.hpp"
+// #include "src/decoders/waas/api/framer.hpp"
```
jonathanmcdermid, line 55: delete

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+#include <iostream>
+
+namespace novatel::edie {
+
+enum class FRAMER_ID
+{
+    AUTOMOTIVE,
+    NOVATEL,
+    NMEA,
+    UNKNOWN
```
jonathanmcdermid, line 71: we're fine having this in the public repo?

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -41,53 +43,67 @@ class FramerBase
...
     //----------------------------------------------------------------------------
-    //! \brief A destructor for the FramerBase class.
+    //! \brief A constructor for the FramerBase class.
+    //
+    //! \param[in] strLoggerName_ String to name the internal logger.
+    //---------------------------------------------------------------------------
+    FramerBase(const std::string& strLoggerName_) : pclMyLogger(Logger::RegisterLogger(strLoggerName_))
+    {
+        pclMyCircularDataBuffer->Clear();
+        pclMyLogger->debug("Framer initialized without Circular Buffer");
```
jonathanmcdermid, line 97: should this be info rather than debug?

include/novatel_edie/common/framer_manager.hpp
rldoris: No this probably shouldn't be there. The system will have to know about types without being enums. Potentially could use the hash system like what we do with the field types.

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    auto it = std::min_element(framerRegistry.begin(), framerRegistry.end(), [](const FramerElement& a, const FramerElement& b) {
+        return a.framer->uiMyFrameBufferOffset < b.framer->uiMyFrameBufferOffset;
+    });
+
+    if (it != framerRegistry.end())
+    {
+        FramerElement earliestFramerElement = std::move(*it);
+        framerRegistry.erase(it);
+        framerRegistry.push_front(std::move(earliestFramerElement));
```
jonathanmcdermid, line 52: I think splice works here and removes move semantics

```suggestion
        framerRegistry.splice(framerRegistry.begin(), framerRegistry, it);
```

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (eActiveFramerId_ == FRAMER_ID::UNKNOWN)
+    {
+        ResetAllMetaDataStates();
+        for (auto& [framerId, framer, metadata, offset] : framerRegistry) { framer->FindNextSyncByte(uiFrameBufferSize_); }
+    }
+
+    // A Framer Found A Sync Byte
+    DisplayFramerStack();
+    // TODO rename this to sort possibly (sort, prioritize, etc)
+    ReorderFramers();
```
jonathanmcdermid, line 150: sort makes sense

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (framerRegistry.front().framer->eMyCurrentFramerStatus == STATUS::INCOMPLETE) { return STATUS::INCOMPLETE; }
+
+    eActiveFramerId_ = framerRegistry.front().framerId;
+
+    if (framerRegistry.front().framer->uiMyFrameBufferOffset > 0 &&
+        (framerRegistry.front().framer->eMyCurrentFramerStatus != STATUS::INCOMPLETE ||
+         framerRegistry.front().framer->eMyCurrentFramerStatus != STATUS::INCOMPLETE_MORE_DATA))
+    {
+
+        HandleUnknownBytes(pucFrameBuffer_, framerRegistry.front().framer->uiMyFrameBufferOffset);
```
jonathanmcdermid, line 163:
```suggestion
        HandleUnknownBytes(pucFrameBuffer_, framerRegistry.front().framer->uiMyFrameBufferOffset);
```

src/decoders/oem/src/framer.cpp
```diff
@@ -293,18 +425,20 @@ Framer::GetFrame(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSiz
                     uiMyByteCount = OEM4_SHORT_BINARY_SYNC_LENGTH;
                     uiMyExpectedPayloadLength = 0;
                     uiMyExpectedMessageLength = 0;
-                    ResetState();
+                    stMetaData.eFormat = HEADER_FORMAT::UNKNOWN;
+                    stMetaData.uiLength = uiMyByteCount;
+                    uiMyFrameBufferOffset = uiMyByteCount;
+                    return STATUS::UNKNOWN;
                     break;
```
jonathanmcdermid, line 432: unnecessary break after return

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+//
+// #include "src/decoders/waas/api/common.hpp"
+// #include "src/decoders/waas/api/framer.hpp"
+
+//----------------------------------------------------------------------------
+//! \brief A constructor for the FramerBase class.
+//
+//! \param[in] strLoggerName_ String to name the internal logger.
+//----------------------------------------------------------------------------
+#include <iostream>
```
jonathanmcdermid, line 62: put with other includes

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+// Forward Declarations of Framers
+// class novatel::edie::oem::Framer;
+
+class FramerManager
+{
+  private:
+    std::list<FRAMER_ID> userFramers;
+    FramerManager();
+    FramerManager(const FramerManager&) = delete;
+    FramerManager& operator=(const FramerManager&) = delete;
```
jonathanmcdermid, line 96: delete move constructor and move operator= as well if you are trying to disable copies

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+};
+
+// Forward Declarations of Framers
+// class novatel::edie::oem::Framer;
+
+class FramerManager
+{
+  private:
+    std::list<FRAMER_ID> userFramers;
```
jonathanmcdermid, line 93: put this member with the other members

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+
+    void ResetInactiveMetaDataStates(const FRAMER_ID& activeFramer_);
+
+  protected:
+    // TODO delete this
+    void DisplayFramerStack();
+
+  public:
+    std::deque<FramerElement> framerRegistry;
+    void RegisterFramer(const FRAMER_ID framerId_, std::unique_ptr<FramerBase>, std::unique_ptr<MetaDataBase>);
```
jonathanmcdermid, line 115: const in declaration is unnecessary

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+    //---------------------------------------------------------------------------
+    //! \brief Get the MetaData for a specific framer.
+    //
+    //! \param[in] framerId_ The ID of the framer to get the MetaData for.
+    //! \return A pointer to the MetaData for the specified framer.
+    //! \return nullptr if the framer is not found.
+    //---------------------------------------------------------------------------
+    MetaDataBase* GetMetaData(const FRAMER_ID framerId_);
```
jonathanmcdermid, line 123: const in declaration is unnecessary

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,240 @@
...
+    ~FramerManager() = default;
+
+    //----------------------------------------------------------------------------
+    //! \brief Get the pointer to a FramerElement Structure for specified framer.
+    //
+    //! \param[in] framerId_ The ID of the framer to get the FramerElement for.
+    //
+    //! \return A pointer to the FramerElement for the specified framer.
+    //---------------------------------------------------------------------------
+    FramerElement* GetFramerElement(const FRAMER_ID framerId_);
```
jonathanmcdermid, line 172: const unnecessary

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -41,53 +43,67 @@ class FramerBase
...
 
-  public:
     //----------------------------------------------------------------------------
     //! \brief A constructor for the FramerBase class.
     //
     //! \param[in] strLoggerName_ String to name the internal logger.
+    //! \param[in] circularBuffer_ Shared pointer to a CircularBuffer object.
     //----------------------------------------------------------------------------
-    FramerBase(const std::string& strLoggerName_) : pclMyLogger(Logger::RegisterLogger(strLoggerName_))
+    FramerBase(const std::string& strLoggerName_, std::shared_ptr<CircularBuffer> circularBuffer_)
```
jonathanmcdermid, line 82: make circularBuffer argument a const reference to avoid unnecessary copies

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -41,53 +43,67 @@ class FramerBase
...
+    }
+
     //----------------------------------------------------------------------------
-    virtual ~FramerBase() = default;
+    //! \brief Find the next sync byte in the circular buffer.
+    //! \param[in] pucFrameBuffer_ The buffer to search for the next sync byte.
+    //! \param[in] uiFrameBufferSize_ The length of pucFrameBuffer_.
+    //
+    //! \return SYNC_BYTE_FOUND if a sync byte is found.
+    virtual void FindNextSyncByte(const uint32_t uiFrameBufferSize_) = 0;
```
jonathanmcdermid, line 106: const unnecessary, function comment description inaccurate.

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -150,23 +163,48 @@ class FramerBase
...
+    //----------------------------------------------------------------------------
+    [[nodiscard]] virtual STATUS GetFrame(unsigned char* pucFrameBuffer_, uint32_t uiFrameBufferSize_, MetaDataBase& stMetaData_) = 0;
+
+    //----------------------------------------------------------------------------
+    //! \brief A destructor for the FramerBase class.
+    //----------------------------------------------------------------------------
+    virtual ~FramerBase() = default;
+};
+} // namespace novatel::edie
 #endif // FRAMER_HPP
```
jonathanmcdermid, line 210:
```suggestion
};

} // namespace novatel::edie

#endif // FRAMER_HPP
```

include/novatel_edie/decoders/oem/framer.hpp
```diff
@@ -57,11 +55,35 @@ class Framer : public FramerBase
...
     //----------------------------------------------------------------------------
     Framer();
 
+    //----------------------------------------------------------------------------
+    //! \brief Find the next sync byte in the circular buffer.
+    //! \param[in] pucFrameBuffer_ The buffer to search for the next sync byte.
+    //! \param[in] uiFrameBufferSize_ The length of pucFrameBuffer_.
+    //! \return The offset of the next sync byte. | -1 if no sync byte is found within the buffer.
+    //---------------------------------------------------------------------------
+    void FindNextSyncByte(const uint32_t uiFrameBufferSize_) override;
```
jonathanmcdermid, line 85: comment description inaccurate, const unnecessary

include/novatel_edie/decoders/oem/framer.hpp
```diff
@@ -81,7 +103,12 @@ class Framer : public FramerBase
     //!   BUFFER_FULL: pucFrameBuffer_ has no more room for added bytes, according
     //! to the size specified by uiFrameBufferSize_.
     //----------------------------------------------------------------------------
-    [[nodiscard]] STATUS GetFrame(unsigned char* pucFrameBuffer_, uint32_t uiFrameBufferSize_, MetaDataStruct& stMetaData_);
+    [[nodiscard]] STATUS GetFrame(unsigned char* pucFrameBuffer_, uint32_t uiFrameBufferSize_, MetaDataBase& stMetaData_) override;
+
+    //----------------------------------------------------------------------------
+    //! \brief A destructor for the Framer class.
+    //----------------------------------------------------------------------------
+    ~Framer() = default;
```
jonathanmcdermid, line 111: mark as override when you override a base destructor

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,221 @@
...
+MetaDataBase* FramerManager::GetMetaData(const FRAMER_ID framerId_)
+{
+    for (FramerElement& element : framerRegistry)
+    {
+        if (element.framerId == framerId_) { return element.metadata.get(); }
+    }
+    return nullptr;
+}
+
+void FramerManager::DisplayFramerStack()
```
jonathanmcdermid, line 65: make this function const

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (it_null_provided != framerRegistry.end())
+    {
+        ResetAllFramerStates();
+        return STATUS::NULL_PROVIDED;
+    }
+
+    return STATUS::INCOMPLETE;
+}
+
+void FramerManager::HandleUnknownBytes(unsigned char* pucBuffer_, const uint32_t& uiUnknownBytes_)
```
jonathanmcdermid, line 217: mark as const

src/decoders/oem/src/framer.cpp
```diff
@@ -27,28 +27,40 @@
...
-Framer::Framer() : FramerBase("novatel_framer") {}
+Framer::Framer(std::shared_ptr<CircularBuffer> circularBuffer) : FramerBase("novatel_framer", circularBuffer)
+{
+    // This static instance persists for the lifetime of the program even after leaving scope of this function
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    auto it_registered = std::find_if(clMyFramerManager.framerRegistry.begin(), clMyFramerManager.framerRegistry.end(),
+                                      [](const FramerElement& element) { return element.framerId == FRAMER_ID::NOVATEL; });
+    if (it_registered != clMyFramerManager.framerRegistry.end()) { throw std::runtime_error("Framer already registered"); }
+
+    clMyFramerManager.RegisterFramer(FRAMER_ID::NOVATEL, std::unique_ptr<Framer>(this), std::make_unique<MetaDataStruct>());
```
jonathanmcdermid, line 44: std::unique_ptr<Framer>(this) causes double deletion

src/decoders/oem/src/framer.cpp
```diff
@@ -27,28 +27,40 @@
...
+}
+
+Framer::Framer() : Framer::Framer(FramerManager::GetInstance().GetCircularBuffer()) {}
```
jonathanmcdermid, line 47:
```suggestion
Framer::Framer() : Framer(FramerManager::GetInstance().GetCircularBuffer()) {}
```

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+    uiMyFrameBufferOffset = 0;
+    ResetState();
+}
 
-    // Loop buffer to complete NovAtel message
-    while (eMyFrameState != NovAtelFrameState::COMPLETE_MESSAGE)
-    {
-        stMetaData_.bResponse = false;
+// -------------------------------------------------------------------------------------------------------
+void Framer::FindNextSyncByte(const uint32_t uiFrameBufferSize_)
```
jonathanmcdermid, line 111: no default case in this switch

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
-             stMetaData_.eFormat == HEADER_FORMAT::JSON) &&
-            ucDataByte > 127)
-        {
-            stMetaData_.eFormat = HEADER_FORMAT::UNKNOWN;
-            ResetState();
-            uiMyByteCount--;
-        }
-
-        switch (eMyFrameState)
+        switch (localFrameState)
```
jonathanmcdermid, line 136: no default case in this switch

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
             uiMyCalculatedCrc32 = 0;
 
             switch (ucDataByte)
             {
-            case OEM4_BINARY_SYNC1:
-                CalculateCharacterCrc32(uiMyCalculatedCrc32, ucDataByte);
-                eMyFrameState = NovAtelFrameState::WAITING_FOR_BINARY_SYNC2;
-                break;
+            case OEM4_BINARY_SYNC1: localFrameState = NovAtelFrameState::WAITING_FOR_BINARY_SYNC2; break;
             case OEM4_ASCII_SYNC:
```
jonathanmcdermid, line 144: use [[fallthrough]] attribute for cases with identical behaviour

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
-                stMetaData_.eFormat = HEADER_FORMAT::UNKNOWN;
-                ResetState();
-                uiMyByteCount--;
+                //   If we get to a WAITING_FOR_ABB_ASCII_SYNC2, it means we got the first byte of the header
+                //   and if it's not immediately followed by sync2, it's not a valid header -> therefore unknown bytes of offset = count
+                uiMyFrameBufferOffset = uiMyByteCount;
+                eMyCurrentFramerStatus = STATUS::UNKNOWN;
+                return;
+            }
+            break;
```
jonathanmcdermid, line 245: redundant break

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
-            if (ucDataByte != OEM4_ABBREV_ASCII_SEPARATOR && (isalpha(ucDataByte) != 0))
+            if (ucDataByte != OEM4_ABBREV_ASCII_SEPARATOR && isalpha(ucDataByte))
             {
-                eMyFrameState = NovAtelFrameState::WAITING_FOR_ABB_ASCII_HEADER;
+                uiMyByteCount = uiMyByteCount - 2;
+                uiMyFrameBufferOffset = uiMyByteCount;
+                eMyCurrentFramerStatus = STATUS::SYNC_BYTES_FOUND;
+                return;
             }
             else
```
jonathanmcdermid, line 237: redundant else

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+            break;
+        }
+    }
+}
+
+// -------------------------------------------------------------------------------------------------------
+STATUS
+Framer::GetFrame(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSize_, MetaDataBase& stMetaData_)
+{
+    MetaDataStruct& stMetaData = dynamic_cast<MetaDataStruct&>(stMetaData_);
```
jonathanmcdermid, line 254:
```suggestion
    auto& stMetaData = dynamic_cast<MetaDataStruct&>(stMetaData_);
```

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +73,296 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+                eMyFrameState = NovAtelFrameState::WAITING_FOR_BINARY_HEADER;
+            }
+            if ((*pclMyCircularDataBuffer)[uiMyByteCount + 2] == OEM4_SHORT_BINARY_SYNC3)
+            {
+                stMetaData.eFormat = HEADER_FORMAT::SHORT_BINARY;
+                eMyFrameState = NovAtelFrameState::WAITING_FOR_SHORT_BINARY_HEADER;
             }
+            for (auto i = 0; i < 3; i++)
+            {
+                const unsigned char ucDataByte = (*pclMyCircularDataBuffer)[uiMyByteCount++];
```
jonathanmcdermid, line 282: we already have a stack variable with this name

src/decoders/oem/src/header_decoder.cpp
```diff
@@ -131,26 +131,27 @@ void HeaderDecoder::DecodeJsonHeader(json clJsonHeader_, IntermediateHeader& stI
 
 // -------------------------------------------------------------------------------------------------------
 STATUS
-HeaderDecoder::Decode(const unsigned char* pucLogBuf_, IntermediateHeader& stInterHeader_, MetaDataStruct& stMetaData_) const
+HeaderDecoder::Decode(const unsigned char* pucLogBuf_, IntermediateHeader& stInterHeader_, MetaDataBase& stMetaData_) const
 {
+    MetaDataStruct& stMetaData = dynamic_cast<MetaDataStruct&>(stMetaData_);
```
jonathanmcdermid, line 136:
```suggestion
    auto& stMetaData = dynamic_cast<MetaDataStruct&>(stMetaData_);
```

src/decoders/oem/src/rxconfig/rxconfig_handler.cpp
```diff
@@ -83,8 +87,12 @@ RxConfigHandler::Convert(MessageDataStruct& stRxConfigMessageData_, MetaDataStru
 
     unsigned char* pucTempMessagePointer = pcMyFrameBuffer.get();
 
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    FRAMER_ID eActiveFramerId = FRAMER_ID::UNKNOWN;
```
jonathanmcdermid, line 91:
```suggestion
    auto eActiveFramerId = FRAMER_ID::UNKNOWN;
```

src/decoders/oem/src/rxconfig/rxconfig_handler.cpp
```diff
@@ -83,8 +87,12 @@ RxConfigHandler::Convert(MessageDataStruct& stRxConfigMessageData_, MetaDataStru
...
+
     // Get an RXCONFIG log.
-    STATUS eStatus = clMyFramer.GetFrame(pcMyFrameBuffer.get(), uiInternalBufferSize, stRxConfigMetaData_);
+    uint32_t uiFrameBufferOffset = 0;
```
jonathanmcdermid, line 94: unused variable

src/decoders/oem/src/rxconfig/rxconfig_handler.cpp
```diff
@@ -70,7 +70,11 @@ bool RxConfigHandler::IsRxConfigTypeMsg(uint16_t usMessageId_)
 }
 
 // -------------------------------------------------------------------------------------------------------
-uint32_t RxConfigHandler::Write(unsigned char* pucData_, uint32_t uiDataSize_) { return clMyFramer.Write(pucData_, uiDataSize_); }
+uint32_t RxConfigHandler::Write(unsigned char* pucData_, uint32_t uiDataSize_)
```
jonathanmcdermid, line 73: make the write functions static as FramerManager is a singleton

src/decoders/oem/test/oem_test.cpp
```diff
@@ -45,51 +46,64 @@
...
 
     // Per-test-suite setup
     static void SetUpTestSuite()
     {
-        pclMyFramer = std::make_unique<Framer>();
-        pclMyFramer->SetReportUnknownBytes(true);
-        pclMyFramer->SetPayloadOnly(false);
+        FramerManager& clMyFramerManager = FramerManager::GetInstance();
+        clMyFramerManager.SetLoggerLevel(spdlog::level::info);
+        std::cout << "DEBUG:FramerTest Constructed Successfully" << std::endl;
```
jonathanmcdermid, line 62: avoid using standard output in tests

src/decoders/oem/test/oem_test.cpp
```diff
@@ -45,51 +46,64 @@
...
     // Per-test-suite teardown
     static void TearDownTestSuite() { Logger::Shutdown(); }
 
     // Per-test setup
-    void SetUp() override { FlushFramer(); }
+    void SetUp() override
+    {
+        FlushTestFixture();
+        pucMyTestFrameBuffer = std::make_unique<unsigned char[]>(131071); // 128k
+        novatel::edie::FramerManager& clMyFramerManager = FramerManager::GetInstance();
```
jonathanmcdermid, line 75: unnecessary namespace specifier

src/decoders/oem/test/oem_test.cpp
```diff
@@ -45,51 +46,64 @@
...
+        clMyFramerManager.ResetAllFramerStates();
+        clMyFramerManager.ResetAllMetaDataStates();
     }
 
+  public:
+    template <HEADER_FORMAT F, STATUS S> void FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_, FRAMER_ID& id_);
+
+    template <HEADER_FORMAT F, STATUS S> void FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_);
+
     static void WriteFileStreamToFramer(std::string sFilename_)
```
jonathanmcdermid, line 94: make argument const reference

src/decoders/oem/test/oem_test.cpp
```diff
@@ -99,16 +113,34 @@ class FramerTest : public ::testing::Test
...
-        ASSERT_EQ(pclMyFramer->Write(pucBytes_, uiNumBytes_), uiNumBytes_);
+        FramerManager& clMyFramerManager = FramerManager::GetInstance();
+        ASSERT_EQ(clMyFramerManager.Write(pucBytes_, uiNumBytes_), uiNumBytes_);
     }
 
-    static void FlushFramer()
+    void FlushTestFixture()
     {
-        while (pclMyFramer->Flush(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH) > 0) {}
+        uint32_t uiBytes = 0;
```
jonathanmcdermid, line 122: unused variable

src/decoders/oem/test/oem_test.cpp
```diff
@@ -99,16 +113,34 @@ class FramerTest : public ::testing::Test
 
     static void WriteBytesToFramer(const unsigned char* pucBytes_, uint32_t uiNumBytes_)
     {
-        ASSERT_EQ(pclMyFramer->Write(pucBytes_, uiNumBytes_), uiNumBytes_);
+        FramerManager& clMyFramerManager = FramerManager::GetInstance();
+        ASSERT_EQ(clMyFramerManager.Write(pucBytes_, uiNumBytes_), uiNumBytes_);
     }
 
-    static void FlushFramer()
+    void FlushTestFixture()
```
jonathanmcdermid, line 120: static

src/decoders/oem/test/oem_test.cpp
```diff
@@ -99,16 +113,34 @@ class FramerTest : public ::testing::Test
...
-        while (pclMyFramer->Flush(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH) > 0) {}
+        uint32_t uiBytes = 0;
+        FramerManager& clMyFramerManager = FramerManager::GetInstance();
+        while (clMyFramerManager.Flush(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH) > 0) {}
+        pucMyTestFrameBuffer.reset();
     }
 };
 
-std::unique_ptr<Framer> FramerTest::pclMyFramer = nullptr;
+template <HEADER_FORMAT F, STATUS S> void FramerTest::FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_, FRAMER_ID& id_)
```
jonathanmcdermid, line 129: static

src/decoders/oem/test/oem_test.cpp
```diff
@@ -99,16 +113,34 @@ class FramerTest : public ::testing::Test
...
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    MetaDataStruct stExpectedMetaData(F, uiLength_);
+    ASSERT_EQ(S, clMyFramerManager.GetFrame(pucMyTestFrameBuffer.get(), uiFrameLength_, id_));
+    MetaDataStruct* stTestMetaData = dynamic_cast<MetaDataStruct*>(clMyFramerManager.GetMetaData(FRAMER_ID::NOVATEL));
+    ASSERT_EQ(*stTestMetaData, stExpectedMetaData);
+}
+
+template <HEADER_FORMAT F, STATUS S> void FramerTest::FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_)
+{
+    FRAMER_ID id = FRAMER_ID::UNKNOWN;
```
jonathanmcdermid, line 140:
```suggestion
    auto id = FRAMER_ID::UNKNOWN;
```

src/decoders/oem/test/oem_test.cpp
```diff
@@ -173,62 +206,64 @@ TEST_F(FramerTest, ASCII_INADEQUATE_BUFFER)
...
+//    stExpectedMetaData.uiLength = uiLogSize;
+//    ASSERT_EQ(STATUS::SUCCESS, clMyFramerManager.GetFrame(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH, FRAMER_ID::UNKNOWN));
+//    ASSERT_EQ(stTestMetaData, stExpectedMetaData);
+//}
 
 TEST_F(FramerTest, ASCII_SEGMENTED)
 {
     constexpr unsigned char aucData[] = "#BESTPOSA,COM1,0,83.5,FINESTEERING,2163,329760.000,02400000,b1f6,65535;SOL_COMPUTED,SINGLE,51.15043874397,-114.03066788586,1097.6822,-17.0000,WGS84,1.3648,1.1806,3.1112,\"\",0.000,0.000,18,18,18,0,00,02,11,01*c3194e35\r\n";
     uint32_t uiBytesWritten = 0;
+    FRAMER_ID id = FRAMER_ID::UNKNOWN;
```
jonathanmcdermid, line 245: use auto throughout file for this scenario

src/decoders/oem/test/oem_test.cpp
```diff
@@ -249,52 +286,55 @@ TEST_F(FramerTest, ABBREV_ASCII_SEGMENTED)
     uint32_t uiLogSize = sizeof(aucData) - 1 - 6; // Remove the [ICOM] from the log size
     uint32_t uiBytesWritten = 0;
     MetaDataStruct stExpectedFrameData;
-    MetaDataStruct stTestMetaData;
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    MetaDataStruct* stTestMetaData = dynamic_cast<MetaDataStruct*>(clMyFramerManager.GetMetaData(FRAMER_ID::NOVATEL));
```
jonathanmcdermid, line 290: auto

src/decoders/oem/test/oem_test.cpp
```diff
@@ -670,122 +729,125 @@ TEST_F(FramerTest, SHORT_BINARY_TRICK)
...
-}
-
-TEST_F(FramerTest, NMEA_INADEQUATE_BUFFER)
-{
-    constexpr unsigned char aucData[] = "$GPALM,30,01,01,2029,00,4310,7b,145f,fd44,a10ce4,1c5b11,0b399f,2bc421,f80,ffe*29\r\n";
-    WriteBytesToFramer(aucData, sizeof(aucData) - 1);
-    FramerHelper<HEADER_FORMAT::NMEA, STATUS::BUFFER_FULL>(sizeof(aucData) - 1, sizeof(aucData) - 2);
-    FramerHelper<HEADER_FORMAT::NMEA, STATUS::SUCCESS>(sizeof(aucData) - 1, sizeof(aucData) - 1);
-}
-
```
jonathanmcdermid, line 722: remove all nmea framer unit tests

src/decoders/oem/test/proprietary_test.cpp
```diff
@@ -36,57 +37,94 @@ using namespace novatel::edie::oem;
...
-        while (pclMyFramer->Flush(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH) > 0) {}
+        uint32_t uiBytes = 0;
+        FramerManager& clMyFramerManager = FramerManager::GetInstance();
+        while (clMyFramerManager.Flush(pucMyTestFrameBuffer.get(), MAX_ASCII_MESSAGE_LENGTH) > 0) {}
+        pucMyTestFrameBuffer.reset();
     }
 };
 
-std::unique_ptr<Framer> ProprietaryFramerTest::pclMyFramer = nullptr;
+template <HEADER_FORMAT F, STATUS S> void ProprietaryFramerTest::FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_, FRAMER_ID& id_)
```
jonathanmcdermid, line 113: static

src/decoders/oem/test/proprietary_test.cpp
```diff
@@ -36,57 +37,94 @@ using namespace novatel::edie::oem;
...
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    MetaDataStruct stExpectedMetaData(F, uiLength_);
+    ASSERT_EQ(S, clMyFramerManager.GetFrame(pucMyTestFrameBuffer.get(), uiFrameLength_, id_));
+    MetaDataStruct* stTestMetaData = dynamic_cast<MetaDataStruct*>(clMyFramerManager.GetMetaData(FRAMER_ID::NOVATEL));
+    ASSERT_EQ(*stTestMetaData, stExpectedMetaData);
+}
+
+template <HEADER_FORMAT F, STATUS S> void ProprietaryFramerTest::FramerHelper(uint32_t uiLength_, uint32_t uiFrameLength_)
+{
+    FRAMER_ID id = FRAMER_ID::UNKNOWN;
```
jonathanmcdermid, line 124: auto

include/novatel_edie/common/framer_manager.hpp
zainarnoldhexagon: That's what this code is being explicit about. By having them equal delete, it enforces the singleton pattern, I'm pretty sure.

src/decoders/oem/src/framer.cpp
zainarnoldhexagon: Please take a look at my solution to this in my next commit, I changed it to std::make_unique<Framer>(*this) since I still want a unique pointer to be passed in, but I want to copy the value in.

src/common/src/framer_manager.cpp
zainarnoldhexagon: deque's apparently don't have the splice functionality

src/decoders/oem/src/rxconfig/rxconfig_handler.cpp
zainarnoldhexagon: static in the header now!

include/novatel_edie/common/framer_manager.hpp
jonathanmcdermid: You need to delete the move operator= and constructor too though. Singleton is supposed to remove everything that can cause

src/decoders/oem/src/framer.cpp
```diff
@@ -27,28 +27,40 @@
...
-Framer::Framer() : FramerBase("novatel_framer") {}
+Framer::Framer(std::shared_ptr<CircularBuffer> circularBuffer) : FramerBase("novatel_framer", circularBuffer)
+{
+    // This static instance persists for the lifetime of the program even after leaving scope of this function
+    FramerManager& clMyFramerManager = FramerManager::GetInstance();
+    auto it_registered = std::find_if(clMyFramerManager.framerRegistry.begin(), clMyFramerManager.framerRegistry.end(),
+                                      [](const FramerElement& element) { return element.framerId == FRAMER_ID::NOVATEL; });
+    if (it_registered != clMyFramerManager.framerRegistry.end()) { throw std::runtime_error("Framer already registered"); }
+
+    clMyFramerManager.RegisterFramer(FRAMER_ID::NOVATEL, std::make_unique<Framer>(*this), std::make_unique<MetaDataStruct>());
```
jonathanmcdermid, line 44: constructing a Framer in the Framer constructor will lead to infinite recursion

include/novatel_edie/common/framer_manager.hpp
zainarnoldhexagon: Resolved in the newest version!

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,231 @@
...
+// class novatel::edie::oem::Framer;
+
+class FramerManager
+{
+  private:
+    FramerManager();
+
+    //! NOTE: disable copies and moves.
+    FramerManager(const FramerManager&) = delete;
+    FramerManager(const FramerManager&&) = delete;
```
jonathanmcdermid, line 83: forgot to mention, on these, the argument shouldnt be const because move constructors and operator= functions promise to destroy the old object

src/decoders/oem/src/framer.cpp
```diff
@@ -27,28 +27,43 @@
...
+    if (it_exists != clMyFramerManager.idMap.end())
+    {
+        auto it_registered =
+            std::find_if(clMyFramerManager.framerRegistry.begin(), clMyFramerManager.framerRegistry.end(),
+                         [&clMyFramerManager](const FramerElement& element) { return element.framerId == clMyFramerManager.idMap["NOVATEL"]; });
+        if (it_registered != clMyFramerManager.framerRegistry.end()) { throw std::runtime_error("Framer already registered"); }
+    }
+}
+
+Framer::Framer() : Framer(FramerManager::GetInstance().GetCircularBuffer()) {}
```
jonathanmcdermid, line 50: if the argument is always the same (circular buffer from singleton) get rid of the param and just do it locally in the function

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +76,285 @@ bool Framer::IsAbbrevAsciiResponse() const
...
-
-            HandleUnknownBytes(pucFrameBuffer_, uiMyByteCount);
-            return STATUS::UNKNOWN;
+            eMyCurrentFramerStatus = STATUS::UNKNOWN;
+            return;
         }
 
-        const unsigned char ucDataByte = clMyCircularDataBuffer[uiMyByteCount++];
-        stMetaData_.uiLength = uiMyByteCount;
+        const unsigned char ucDataByte = (*pclMyCircularDataBuffer)[uiMyByteCount++];
```
jonathanmcdermid, line 141: is this variable necessary? I feel like the code would be clearer if it was removed

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +76,285 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+}
+
+// -------------------------------------------------------------------------------------------------------
+STATUS
+Framer::GetFrame(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSize_, MetaDataBase& stMetaData_)
+{
+    auto& stMetaData = dynamic_cast<MetaDataStruct&>(stMetaData_);
+
+    // Trust Relationship: FindNextSyncByte leads to this function, so we can trust that the first byte is the start of a sync byte
+    unsigned char ucDataByte = (*pclMyCircularDataBuffer)[uiMyByteCount];
```
jonathanmcdermid, line 249: is this variable necessary? I feel like the code would be clearer if it was removed

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +76,285 @@ bool Framer::IsAbbrevAsciiResponse() const
...
-            else
+            if ((*pclMyCircularDataBuffer)[uiMyByteCount + 2] == OEM4_BINARY_SYNC3)
             {
-                stMetaData_.eFormat = HEADER_FORMAT::UNKNOWN;
-                ResetState();
-                uiMyByteCount--;
+                if (stMetaData.eFormat != HEADER_FORMAT::PROPRIETARY_BINARY) { stMetaData.eFormat = HEADER_FORMAT::BINARY; }
+                eMyFrameState = NovAtelFrameState::WAITING_FOR_BINARY_HEADER;
             }
+            if ((*pclMyCircularDataBuffer)[uiMyByteCount + 2] == OEM4_SHORT_BINARY_SYNC3)
```
jonathanmcdermid, line 267: else if

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +76,285 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+            {
+                stMetaData.eFormat = HEADER_FORMAT::SHORT_BINARY;
+                eMyFrameState = NovAtelFrameState::WAITING_FOR_SHORT_BINARY_HEADER;
+            }
+            for (auto i = 0; i < 3; i++)
+            {
+                ucDataByte = (*pclMyCircularDataBuffer)[uiMyByteCount++];
```
jonathanmcdermid, line 274: pass directly to the function

src/decoders/oem/src/framer.cpp
```diff
@@ -61,175 +76,285 @@ bool Framer::IsAbbrevAsciiResponse() const
...
+                return STATUS::INCOMPLETE;
+            }
+
+            stMetaData.eFormat = HEADER_FORMAT::UNKNOWN;
+            stMetaData.uiLength = uiMyByteCount;
+
+            if (uiMyByteCount == 0) { return STATUS::BUFFER_EMPTY; }
+
+            uiMyFrameBufferOffset = uiMyByteCount;
+            stMetaData.uiLength = uiMyByteCount;
```
jonathanmcdermid, line 335: repeat line?

src/decoders/oem/src/framer.cpp
```diff
@@ -508,9 +642,9 @@ Framer::GetFrame(unsigned char* pucFrameBuffer_, const uint32_t uiFrameBufferSiz
             {
                 uiMyByteCount--; // rewind back to delimiter before copying
                 char acCrc[OEM4_ASCII_CRC_LENGTH + 1];
```
jonathanmcdermid, line 644: replace this with string_view logic, remove unnecessary copies

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -148,28 +154,32 @@ int main(int argc, char* argv[])
 
     tStart = std::chrono::high_resolution_clock::now();
 
+    auto eActiveFramerId = clFramerManager.idMap["UKNOWN"];
```
rldoris, line 157: Why is this required at the user level?

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -148,28 +154,32 @@ int main(int argc, char* argv[])
...
         while (eFramerStatus != STATUS::BUFFER_EMPTY && eFramerStatus != STATUS::INCOMPLETE)
         {
             unsigned char* pucFrameBuffer = acFrameBuffer;
-            eFramerStatus = clFramer.GetFrame(pucFrameBuffer, sizeof(acFrameBuffer), stMetaData);
+            eFramerStatus = clFramerManager.GetFrame(pucFrameBuffer, sizeof(acFrameBuffer), eActiveFramerId);
 
             if (eFramerStatus == STATUS::SUCCESS)
             {
+                // OEM logs are the only ones allowed in this example
+                auto& stMetaData = reinterpret_cast<MetaDataStruct&>(*(clFramerManager.framerRegistry.front().metadata));
```
rldoris, line 174: Unsure what this code is doing, but it's calling front() which looks like it should be internal code.

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -148,28 +154,32 @@ int main(int argc, char* argv[])
...
                 if (stMetaData.bResponse)
                 {
                     unknownOfs.write(reinterpret_cast<char*>(pucFrameBuffer), stMetaData.uiLength);
                     continue;
                 }
 
                 pucFrameBuffer[stMetaData.uiLength] = '\0';
-                pclLogger->info("Framed: {}", reinterpret_cast<char*>(pucFrameBuffer));
+                // pclLogger->debug("Framed: {}", reinterpret_cast<char*>(pucFrameBuffer));
```
rldoris, line 182: Remove comments if they aren't needed.

include/novatel_edie/decoders/common/common.hpp
```diff
@@ -255,8 +257,14 @@ struct EnumDefinition;
 //! \brief Base class for all metadata. Contains data about the data such as header format,
 //! GPS week and milliseconds, and message ID.
 //============================================================================
-struct MetaDataBase
+class MetaDataBase
```
rldoris, line 260: Why change this to a class?

python/bindings/framer.cpp
```diff
@@ -21,27 +21,19 @@ void init_novatel_framer(nb::module_& m)
         .def_prop_ro("bytes_available_in_buffer", [](const oem::Framer& framer) { return framer.GetBytesAvailableInBuffer(); })
         .def(
             "get_frame",
-            [](oem::Framer& framer, oem::MetaDataStruct* metadata_ptr, uint32_t buffer_size) {
+            [](oem::Framer& self, uint32_t buffer_size, oem::MetaDataBase* metadatabase_ptr) {
```
rldoris, line 24: Why change the order?

examples/novatel/converter_components/converter_components.cpp
zainarnoldhexagon: You're right, it should be stored within the framer manager

python/bindings/framer.cpp
zainarnoldhexagon: idk who changed the order while I was working on this, but I've been working with that order in my branch the whole time. I'm just changing it back

include/novatel_edie/decoders/common/common.hpp
zainarnoldhexagon: No reason, was an artifact of the rebase

python/bindings/framer.cpp
rldoris: Was this coming from the rebase and you do it?

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -148,28 +154,31 @@ int main(int argc, char* argv[])
...
         while (eFramerStatus != STATUS::BUFFER_EMPTY && eFramerStatus != STATUS::INCOMPLETE)
         {
             unsigned char* pucFrameBuffer = acFrameBuffer;
-            eFramerStatus = clFramer.GetFrame(pucFrameBuffer, sizeof(acFrameBuffer), stMetaData);
+            eFramerStatus = clFramerManager.GetFrame(pucFrameBuffer, sizeof(acFrameBuffer));
 
             if (eFramerStatus == STATUS::SUCCESS)
             {
+                // OEM logs are the only ones allowed in this example
+                auto& stMetaData = reinterpret_cast<MetaDataStruct&>(*(clFramerManager.GetFramerElement(clFramerManager.ActiveFramerId()))->metadata);
```
rldoris, line 174: This code still confuses me. Having to pass in the activeframerid seems like a problem or just confusing interface. I think there should be a call to the framermanager to get the metadata of the active/successful framer.

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -96,13 +98,18 @@ int main(int argc, char* argv[])
...
     // Set up the EDIE components
-    Framer clFramer;
-    clFramer.SetLoggerLevel(spdlog::level::debug);
-    Logger::AddConsoleLogging(clFramer.GetLogger());
-    Logger::AddRotatingFileLogger(clFramer.GetLogger());
-    clFramer.SetReportUnknownBytes(true);
-    clFramer.SetPayloadOnly(false);
-    clFramer.SetFrameJson(false);
+    // get reference to static singleton instance of FramerManager
+    novatel::edie::FramerManager& clFramerManager = FramerManager::GetInstance();
```
jonathanmcdermid, line 102: unnecessary namespace

examples/novatel/converter_components/converter_components.cpp
```diff
@@ -193,33 +202,40 @@ int main(int argc, char* argv[])
...
                             stMessageData.pucMessage[stMessageData.uiMessageLength] = '\0';
-                            pclLogger->info("Encoded: ({}) {}", stMessageData.uiMessageLength, reinterpret_cast<char*>(pucEncodedMessageBuffer));
+                            pclLogger->debug("Encoded: ({}) {}", stMessageData.uiMessageLength, reinterpret_cast<char*>(pucEncodedMessageBuffer));
                         }
                         else
                         {
                             unknownOfs.write(reinterpret_cast<char*>(pucFrameBuffer), uiBodyLength);
                             pclLogger->warn("Encoder returned with status code {}", eEncoderStatus);
                         }
+                        clFramerManager.ResetActiveFramerId();
```
jonathanmcdermid, line 212: if we do this in all the branches just put it in the common path at the end

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+{
+  private:
+    const std::string framerManagerName;
+    FramerManager();
+
+    //! NOTE: disable copies and moves.
+    FramerManager(FramerManager&) = delete;
+    FramerManager(FramerManager&&) = delete;
+    FramerManager& operator=(const FramerManager&) = delete;
+    FramerManager& operator=(const FramerManager&&) = delete;
```
jonathanmcdermid, line 75: this shouldnt be const

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+};
+
+class FramerManager
+{
+  private:
+    const std::string framerManagerName;
+    FramerManager();
+
+    //! NOTE: disable copies and moves.
+    FramerManager(FramerManager&) = delete;
```
jonathanmcdermid, line 72: this should be const

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+    FramerElement(int framerId_, std::shared_ptr<FramerBase> framer_, std::shared_ptr<MetaDataBase> metadata_, int32_t offset_)
+        : framerId(framerId_), framer(std::move(framer_)), metadata(std::move(metadata_)), offset(offset_)
+    {
+    }
+};
+
+class FramerManager
+{
+  private:
+    const std::string framerManagerName;
```
jonathanmcdermid, line 68: group all the members together

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+    //---------------------------------------------------------------------------
+    //! \brief Get the MetaData for a specific framer.
+    //
+    //! \param[in] framerName The name of the framer to get the MetaData for.
+    //! \return A pointer to the MetaData for the specified framer.
+    //! \return nullptr if the framer is not found.
+    //---------------------------------------------------------------------------
+    MetaDataBase* GetMetaData(std::string sFramerName)
+    {
+        if (idMap.find(sFramerName) == idMap.end()) { return nullptr; }
```
jonathanmcdermid, line 125: inefficient to do 2 lookups like this

`auto it = idMap.find(sFramerName);
if (it == idMap.end()) { 
    return nullptr; 
}
return GetMetaData(it->second);`

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+    //! \brief Reset the MetaData for the requested framer.
+    //---------------------------------------------------------------------------
+    void ResetMetaData(int iFramerId_)
+    {
+        for (FramerElement& element : framerRegistry)
+        {
+            if (element.framerId == iFramerId_)
+            {
+                element.metadata.reset();
+                break;
```
jonathanmcdermid, line 147: return

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+    //! \brief Set the active Framer ID
+    //---------------------------------------------------------------------------
+    bool SetActiveFramerId(int iFramerId)
+    {
+        for (const auto& pair : idMap)
+        {
+            if (pair.second == iFramerId) { break; }
+        }
+        iActiveFramerId = iFramerId;
+        return (iActiveFramerId == iFramerId);
```
jonathanmcdermid, line 167: always true?

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    //----------------------------------------------------------------------------
+    //! \brief Set the active Framer ID
+    //---------------------------------------------------------------------------
+    bool SetActiveFramerId(std::string sFramerName)
+    {
+        if (idMap.find(sFramerName) == idMap.end()) { return false; }
+        iActiveFramerId = idMap[sFramerName];
```
jonathanmcdermid, line 176: same double lookup as before, store the lookup 

auto it = idMap.find(sFramerName);
if (it == idMap.end()) .....

include/novatel_edie/common/framer_manager.hpp
```diff
@@ -0,0 +1,334 @@
...
+        return true;
+    }
+
+    int GetActiveFramerId() { return iActiveFramerId; }
```
jonathanmcdermid, line 180: make functions in this file const when applicable

include/novatel_edie/decoders/common/common.hpp
```diff
@@ -267,8 +275,23 @@ struct MetaDataBase
...
+        uiLength = 0;
+        uiBinaryMsgLength = 0;
+        uiHeaderLength = 0;
+        usMessageId = 0;
+        uiMessageCrc = 0;
+        messageName.clear();
+    }
 };
 
+inline MetaDataBase::~MetaDataBase() = default;
```
jonathanmcdermid, line 293: why inline?

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -121,20 +142,15 @@ class FramerBase
...
-    //----------------------------------------------------------------------------
-    void SetReportUnknownBytes(const bool bReportUnknownBytes_) { bMyReportUnknownBytes = bReportUnknownBytes_; }
-
     //----------------------------------------------------------------------------
     //! \brief Get the number of bytes available in the internal circular buffer.
     //
     //! \return The number of bytes available in the internal circular buffer.
     //----------------------------------------------------------------------------
-    [[nodiscard]] uint32_t GetBytesAvailableInBuffer() const { return clMyCircularDataBuffer.GetCapacity() - clMyCircularDataBuffer.GetLength(); }
+    [[nodiscard]] virtual uint32_t GetBytesAvailableInBuffer() const
```
jonathanmcdermid, line 150: why virtual?

include/novatel_edie/decoders/common/framer.hpp
```diff
@@ -145,22 +161,52 @@ class FramerBase
...
+
+    ////----------------------------------------------------------------------------
+    ////! \brief Flush bytes from the internal circular buffer.
+    ////
+    ////! \param[in] pucBuffer_ The buffer to contain the flushed bytes.
+    ////! \param[in] uiBufferSize_ The size of the provided buffer.
+    ////
+    ////! \return The number of bytes flushed from the internal circular buffer.
+    ////----------------------------------------------------------------------------
+    virtual uint32_t Flush(uint32_t uiBufferSize_)
```
jonathanmcdermid, line 177: why virtual?

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+#include "novatel_edie/common/framer_manager.hpp"
+
+using namespace novatel::edie;
+
+FramerManager::FramerManager()
+    : framerManagerName("FramerManager"), pclMyCircularDataBuffer(std::make_shared<CircularBuffer>()), iActiveFramerId(-1),
+      pclMyLogger(Logger::RegisterLogger(framerManagerName))
+{
+    pclMyCircularDataBuffer->Clear();
```
jonathanmcdermid, line 35: why is this necessary?

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+void FramerManager::RegisterFramer(std::string framerName, std::shared_ptr<FramerBase> framer_, std::shared_ptr<MetaDataBase> metadata_)
+{
+
+    int framerId_ = -1;
+    auto it = idMap.find(framerName);
+    if (it == idMap.end())
+    {
+        int newId = static_cast<int>(idMap.size());
+        idMap[framerName] = newId;
+        framerId_ = newId;
```
jonathanmcdermid, line 49:
```suggestion
        framerId_ = static_cast<int>(idMap.size());
        idMap[framerName] = framerId_;
```

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+FramerManager::FramerManager()
+    : framerManagerName("FramerManager"), pclMyCircularDataBuffer(std::make_shared<CircularBuffer>()), iActiveFramerId(-1),
+      pclMyLogger(Logger::RegisterLogger(framerManagerName))
+{
+    pclMyCircularDataBuffer->Clear();
+    idMap["UNKNOWN"] = 0;
+    pclMyLogger->debug("Framer Manager initialized");
+}
+
+void FramerManager::RegisterFramer(std::string framerName, std::shared_ptr<FramerBase> framer_, std::shared_ptr<MetaDataBase> metadata_)
```
jonathanmcdermid, line 40: unnecessary string copy

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+            {
+                idName = pair.first;
+                break;
+            }
+        }
+        pclMyLogger->debug("Framer: {}, Offset: {}, Status: {}\n", idName, elem.framer->uiMyFrameBufferOffset, elem.framer->eMyCurrentFramerStatus);
+    }
+}
+
+void FramerManager::ResetInactiveFramerStates(const int& iActiveFramerId_)
```
jonathanmcdermid, line 89: const int& param is unnecessary

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+        if (framerId != iActiveFramerId_)
+        {
+            framer->ResetStateAndByteCount();
+            offset = 0;
+        }
+    }
+}
+
+void FramerManager::ResetAllFramerStates() { ResetInactiveFramerStates(idMap["UNKNOWN"]); }
```
jonathanmcdermid, line 101: just provide the invalid index directly as an rvalue, no map lookup necessary

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+void FramerManager::ResetInactiveMetaDataStates(const int& iActiveFramerId_)
```
jonathanmcdermid, line 103: const int& unnecessary

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    for (auto& [framerId, framer, metadata, offset] : framerRegistry)
+    {
+        if (framerId != iActiveFramerId_) { metadata->ResetMetaData(); }
+    }
+}
+
+void FramerManager::ResetAllMetaDataStates() { ResetInactiveMetaDataStates(idMap["UNKNOWN"]); }
```
jonathanmcdermid, line 111: just directly pass the rvalue of an invalid ID

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include "novatel_edie/common/framer_manager.hpp"
+
+using namespace novatel::edie;
+
+FramerManager::FramerManager()
+    : framerManagerName("FramerManager"), pclMyCircularDataBuffer(std::make_shared<CircularBuffer>()), iActiveFramerId(-1),
+      pclMyLogger(Logger::RegisterLogger(framerManagerName))
+{
+    pclMyCircularDataBuffer->Clear();
+    idMap["UNKNOWN"] = 0;
```
jonathanmcdermid, line 36: Is this necessary?

src/common/src/framer_manager.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    // State Machine:
+    // use framers to identify sync bytes and offsets
+    // once sync bytes are identified, set Active Framer and perform framing using type specific framer, reset inactive framers
+    // upon successful framing of a log, reset Active Framer ID
+
+    // if STATUS is INCOMPLETE or SUCCESS upon entering GetFrame, it is stale
+    if (iActiveFramerId != idMap["UNKNOWN"])
+    {
+        if (framerRegistry.front().framer->eMyCurrentFramerStatus == STATUS::INCOMPLETE)
```
jonathanmcdermid, line 132: instead of calling front() over and over just take a reference to it at the start of the function

PR #87, opened 2025-01-20
Fix example section of readme and delete empty example folder

riley-kinahan:
The current version of the readme has two examples, both of which rely on a `InputFileStream` class which does not exist within the current version of `novatel_edie`. This pull request replaces those examples short explanations and links to working examples from the `./examples/` folder.

It also deletes the `./examples/novatel/rangecmp/` folder which is empty save for a CMake file, and appears to have been superseded by the `./examples/novatel/range_decompressor/` folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,106 +136,62 @@ when bypassing the Parser class.
 
 ## Examples
 
-More examples are provided in the [examples](examples) folder.
+Examples are provided in the [examples](examples) folder.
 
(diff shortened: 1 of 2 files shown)


jonathanmcdermid (approved):
looks good. I'm ambivalent on whether or not we need a description for every example in the readme, but I agree with removing the code examples from it.

Issue #90, opened 2025-01-21
Address Python object representation


riley-kinahan:
The python representation of log message data has some points for improvement:

- **Header/Body/Message data distribution:** Logs instances are currently represented on the Python side as a combination of Header and Message objects. The Message object contains data from the log body, and some supplemental data. It may be more convenient to make a Header and Body object part of a single Message object. This should also eliminate the weirdness of having Message objects nested within Message objects, as Body objects won't contain default fields.

- **Magic method implementation:** Current magic method implementations allow for both attribute-like and dict-like access of fields within a Message object. It's uncommon for data to be accessible in both ways so we should consider dropping one. Implementing a custom `__dir__` method would allow dynamic fields to be directly visible within a debugger, which would make attribute-like access easier.

- **Type info:** It might be possible to incorporate a richer type hierarchy into objects, e.g. `BestposBody`, `RangeBody`, etc. without sacrificing efficiency. This potentially opens the door to better static type-hinting or at least using `isinstance` checks. This should only be included if it can be verified to have a negligible performance effect.

riley-kinahan:
Going to start poking at this, once I get something I'm happy with I'll open a PR for it

valgur:
Shouldn't #76 be merged first?

riley-kinahan:
Yep, I'm just doing some early work based on your branch for now. I believe you should be able to merge #76

PR #92, opened 2025-01-27
Improve Python Object Rep

riley-kinahan:
This merge request addresses #90 in the following ways:
1. Splitting `PyDecodedMessage` into `PyMessage`(`Message`) and `PyMessageBody`(`MessageBody`)
    - `PyMessageBody` contains data about the body of a log.
    - `PyMessage` contains `nb:object` references to the Python representations of a `PyMessageBody`/`MessageBody` and an `IntermediateHeader`/ `Header`
    - `MessageDecoder.decode` returns a `Message` and takes in an additional `Header` parameter
    - `MessageEncoder.encode` takes `Message`(`PyMessage`) instead of a `Header`(`IntermediateHeader`) and a `Message`(`PyDecodedMessage`)
 2. Implementing a `__dir__` method for the new `MessageBody`/`PyMessageBody` class.
     - Adds all field names to the list generated by the default implementation.
 3. Adding subtype information to all message objects.
     - One message type is generated for each message definition, e.g. `RANGE` -> `Message`
     - One message body type is generated for each definition, e.g. `RANGE_Body` -> `MessageBody`
     - Each field array within a message body gets its own definition recursively, e.g. `RANGE_Body_obs_Field` -> `MessageBody`
     - When instantiating an object, the type is chosen based on name. Should be safe from collision as long as no messages names end with `_Body` and subfield names end with `_Field`.
     - Undecodable messages are given `UNKNOWN`/`UNKNOWN_Body` types. Only most recent definition for a message is supported, fields which don't match up will end up with an `UNKNOWN_Body` type.
     - Messages of default database are accessible via `from novatel_edie.message import ...`
     - Type checks of the kind `isinstance(message, RANGE)` are supported. These checks will automatically be enhanced with type hinting by most IDEs once stub file generation scheme is implemented.

Quick Benchmark data:
```
- Decoding 100,000 messages:
    - Current: 25.4s
    - New: 25.8s
- Decoding 100,000 messages and accessing all range observations (nested field conversion)
    - Current: 42.7s
    - New: 41.2s
```


python/bindings/message_decoder.cpp
```diff
@@ -216,13 +252,38 @@ void init_novatel_message_decoder(nb::module_& m)
...
+                PyMessage* message_cinst = nb::inst_ptr<PyMessage>(message_pyinst);
+                new (message_cinst) PyMessage(body_pyinst, header, message_name);
+                nb::inst_mark_ready(message_pyinst);
+
+                return nb::make_tuple(status, message_pyinst);
             },
-            "message_body"_a, "metadata"_a)
-        // For internal testing purposes only
+            "message_body"_a, "decoded_header"_a, "metadata"_a)
         .def(
```
riley-kinahan, line 287: Left these mostly unchanged for now. Feels a little out of place with the new changes so at some point we may want to update the unit tests to be more specific to the Python side of things.

riley-kinahan:
@valgur Looks like I'm not actually able to add you as a reviewer, but feel free to provide any feedback about the changes here

python/bindings/message_decoder.cpp
```diff
@@ -37,7 +39,29 @@ nb::object convert_field(const FieldContainer& field, const PyMessageDatabase::C
...
+                const auto& message_subfield = std::get<std::vector<FieldContainer>>(subfield.fieldValue);
+                new (cinst) PyMessageBody(message_subfield, parent_db, field_name);
+                nb::inst_mark_ready(pyinst);
+                sub_values.push_back(pyinst); 
+            }
+            return nb::cast(sub_values);
+        }
+        else
         {
             // Fixed-length, variable-length and field arrays are stored as a field
```
rldoris, line 66: This comment says that field arrays were being handled by this case already. Was that not happening?

riley-kinahan: Good catch! This comment is outdated and so I'll reword it.

The original implementation was correct. The reason for my change is that once I started adding type information to each element in a field array it became more efficient to process the full array in one go rather than recursing through each element.

python/bindings/message_decoder.cpp
```diff
@@ -81,22 +100,20 @@ nb::object convert_field(const FieldContainer& field, const PyMessageDatabase::C
...
+    : fields(std::move(message_)), parent_db_(std::move(parent_db_)), name(std::move(name_))
+{}
 
-nb::dict& PyDecodedMessage::get_values() const
+nb::dict& PyMessageBody::get_values() const
 {
     if (cached_values_.size() == 0)
     {
-        for (const auto& field : fields) { cached_values_[nb::cast(field.fieldDef->name)] = convert_field(field, parent_db_); }
+        for (const auto& field : fields) { cached_values_[nb::cast(field.fieldDef->name)] = convert_field(field, parent_db_, name); }
```
rldoris, line 111: Where does the name variable come from. I don't see it defined anywhere in this function.

riley-kinahan: Its an attribute of the `PyMessageBody` instance. I can rename it to something like `pytype_name` if you think its unclear.

python/bindings/message_database.cpp
```diff
@@ -222,3 +235,48 @@ inline void PyMessageDatabase::UpdatePythonEnums()
         enums_by_name[enum_name] = enum_type;
     }
 }
+
+void PyMessageDatabase::AddFieldType(std::vector<std::shared_ptr<BaseField>> fields, std::string base_name, nb::handle type_constructor,
+                                     nb::handle type_tuple, nb::handle type_dict)
+{
+    // rescursively add field types for each field array element within the provided vector
```
rldoris, line 242: Why add field type? Could you explain the effect this is having on the Python side?

riley-kinahan: It allows for consistency between type-hinting and instance checks for subfields:

image

image

image

python/bindings/message_database.cpp
```diff
@@ -222,3 +235,48 @@ inline void PyMessageDatabase::UpdatePythonEnums()
...
+            auto* field_array_field = dynamic_cast<FieldArrayField*>(field.get());
+            std::string field_name = base_name + "_" + field_array_field->name + "_Field";
+            nb::object field_type = type_constructor(field_name, type_tuple, type_dict);
+            messages_by_name[field_name] = field_type;
+            AddFieldType(field_array_field->fields, field_name, type_constructor, type_tuple, type_dict);
+        }   
+    }
+}
+
+void PyMessageDatabase::UpdateMessageTypes()
```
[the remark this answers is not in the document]
riley-kinahan: Added a comments to the header file for both the functions that update Python types, which should hopefully add some clarity

python/bindings/message_decoder.cpp
rldoris: Can you access the member using `this->name`? That would be more clear.

python/bindings/py_decoded_message.hpp
```diff
@@ -32,19 +40,27 @@ struct PyDecodedMessage
...
+//============================================================================
+//! \class PyMessage
+//! \brief A python representation for a single log message.
+//! 
+//! Extends PyField with reference to the Python represenation of a Header.
+//============================================================================
+struct PyMessage : public PyField
+{
+  public:
+    nb::object header;
```
valgur, line 60:
```suggestion
    oem::IntermediateHeader header;
```
I think you can avoid a tiny bit of indirection here and embed the header directly inside the PyMessage. Makes it a bit more cache-friendly.

python/bindings/py_decoded_message.hpp
```diff
@@ -18,9 +18,17 @@ struct PyGpsTime
...
+//! \class PyField
+//! \brief A python representation for a single log message or message field.
+//! 
+//! Contains a vector of FieldContainer objects, which behave like attributes 
+//! within the Python API.
+//============================================================================
+struct PyField
 {
-    explicit PyDecodedMessage(std::vector<FieldContainer> message_, const MetaDataStruct& meta_, PyMessageDatabase::ConstPtr parent_db_);
+    std::string name;
```
valgur, line 30: Does the name attribute make sense for fields as well, in addition to messages?

python/test/test_decode_encode.py
```diff
@@ -502,32 +505,33 @@ def test_flat_binary_log_decode_bestpos(helper):
...
-    assert bestpos.num_soln_svs == 30
-    assert bestpos.num_soln_L1_svs == 30
-    assert bestpos.num_soln_multi_svs == 30
-    assert bestpos.extended_solution_status2 == 0x00
-    assert bestpos.ext_sol_stat == 0x06
-    assert bestpos.gal_and_bds_mask == 0x39
-    assert bestpos.gps_and_glo_mask == 0x33
+    assert isinstance(message, BESTPOS)
+    assert message.solution_status == ne.enums.SolStatus.SOL_COMPUTED
+    assert message.position_type == ne.enums.SolType.SINGLE
```
valgur, line 510:
```suggestion
    assert message.solution_status == SolStatus.SOL_COMPUTED
    assert message.position_type == SolType.SINGLE
```
Can make these and other checks more concise now that `from novatel_edie.enums import *` works.

python/bindings/py_decoded_message.hpp
riley-kinahan: It is needed for identifying a few doubly nested fields such as `ITFILTTABLE_filter_coef_status_Field_nf_status_Field`. 

However, this made me realize that it makes sense to expose the `name` attribute for messages (but not fields) within the Python API.

python/bindings/message_decoder.cpp
```diff
@@ -126,31 +148,30 @@ nb::dict PyDecodedMessage::to_dict() const
...
     bool first = true;
     for (const auto& [field_name, value] : get_values())
     {
         if (!first) { repr << ", "; }
         first = false;
         repr << nb::str("{}={!r}").format(field_name, value).c_str();
     }
-    repr << ">";
+    repr << ")";
     return repr.str();
```
rldoris, line 175: Do you think replacing with iterators make for better code?

```
std::stringstream repr;
repr << name << "(";

const auto& values = get_values();
if (!values.empty()) {
    auto it = values.cbegin();
    repr << nb::str("{}={!r}").format(it->first, it->second).c_str();
    for (++it; it != values.cend(); ++it) {
        repr << ", " << nb::str("{}={!r}").format(it->first, it->second).c_str();
    }
}

repr << ")";
return repr.str();
```

If nothing else it removes the `bool first` variable. Blatantly ripped from copilot.

python/bindings/message_decoder.cpp
```diff
@@ -216,13 +262,32 @@ void init_novatel_message_decoder(nb::module_& m)
...
+                {
+                    body_pytype = parent_db->GetMessagesByNameDict().at("UNKNOWN");
+                }
+
+                nb::object body_pyinst = nb::inst_alloc(body_pytype);
+                PyMessage* body_cinst = nb::inst_ptr<PyMessage>(body_pyinst);
+                new (body_cinst) PyMessage(fields, parent_db, message_name, header);
+                nb::inst_mark_ready(body_pyinst);
+
+                //auto message_pyinst = std::make_shared<PyMessage>(body_pyinst, header, message_name);
```
rldoris, line 287: remove comment?

python/bindings/message_decoder.cpp
```diff
@@ -216,13 +262,32 @@ void init_novatel_message_decoder(nb::module_& m)
...
+                nb::handle body_pytype;
+                const std::string message_name = metadata.MessageName();
+
+                try
+                {
+                    body_pytype = parent_db->GetMessagesByNameDict().at(message_name);
+                }
+                catch (const std::out_of_range& e)
+                {
+                    body_pytype = parent_db->GetMessagesByNameDict().at("UNKNOWN");
```
rldoris, line 279: What are the implications of this case? What will the user get a Unknown log back or something else? I'm wondering if we should be handling this differently.

python/bindings/py_decoded_message.hpp
riley-kinahan: Good suggestion! I had to introduce a `PyHeader` subclass of `IntermediateHeader` to get the `Message.to_dict` method to still work, but I actually think this improves the code organization.

python/bindings/message_decoder.cpp
riley-kinahan: I agree this looks better but doesn't appear that nanobind dicts support this kind of iteration

python/bindings/message_decoder.cpp
riley-kinahan: Yes, they will get the `UNKNOWN` type
image

I think it is probably more Pythonic to return `None` so if you agree I'll switch that over.

Also as I look back at this I think the `out_of_range` case should just throw, since an actually `UNKNOWN` message will have the name 'UNKNOWN'. 

Subfields of messages that can't be associated with a type are also returned with an `UNKNOWN` type right now, I think it makes more sense to use a generic `Field` type instead.

python/bindings/message_decoder.cpp
```diff
@@ -185,22 +206,48 @@ void init_novatel_message_decoder(nb::module_& m)
...
-        .def("to_dict", &PyDecodedMessage::to_dict, "Convert the message and its sub-messages into a dict")
-        .def("__getattr__", &PyDecodedMessage::getattr, "field_name"_a)
-        .def("__getitem__", &PyDecodedMessage::getitem, "field_name"_a)
-        .def("__contains__", &PyDecodedMessage::contains, "field_name"_a)
-        .def("__len__", &PyDecodedMessage::len)
-        .def("__repr__", &PyDecodedMessage::repr)
-        .def("__str__", &PyDecodedMessage::repr);
+    nb::class_<PyField>(m, "Field")
+        //.def_prop_ro("_values", &PyField::get_values)
+        //.def_prop_ro("_fields", &PyField::get_fields)
```
rldoris, line 211: Remove comments

python/bindings/message_decoder.cpp
rldoris: In the case above we wouldn't be able to access fields because they couldn't be decoded, but it is a valid log so we could decode the header correct?

What happens with the current version of EDIE? I think we have a STATUS for NO_DEFINITION correct? I like the idea of providing a unknown log with a header to the user, but they need to know that the message definition wasn't found. Another thing to consider is that if binary is decoded then there will be a message_id, but no name and if it's ASCII then name, but no message_id. That inconsistency might be a pain for users to deal with. Might need to think about this.

rldoris:
How are unknown bytes being handled? If the decoder fails on bytes is the user still able to see them (e.g. write to a file or handle some other way)?

rldoris (approved):
I'm reasonably happy with this. In the interest of not having another monster review I say we merge it and can make any needed tweaks with smaller PRs.

valgur:
LGTM as well.

Run clang-format on the codebase before merging, since I think there were some minor inconsistencies.

Issue #93, opened 2025-01-28
Introduce Static Type Hinting for regular classes and default enums/messages


riley-kinahan:
The package does not support static-type hinting. We should make use of [nanobind's tool for generating .pyi stub files](https://nanobind.readthedocs.io/en/latest/typing.html#stub-generation) to enable it.

We should also include static type hints for messages and enums which are part of the default database. Ideally there should also be a way to customize an installation with a different default database, along with different type hints. For this issue a script that does it is sufficient, but eventually we may want to migrate that to its own tool.

The following depicts a goal scenario for how type info should exist within the package:

![Image](https://github.com/user-attachments/assets/20f39f28-e1bb-4a1e-b72f-17528250d199)

*Comparability across databases can be addressed separately*

PR #96, opened 2025-02-05
Add type hinting support

riley-kinahan:
This merge request addresses #93 and is currently blocked by #92 

It introduces the following changes:
1. Renamed 'messages_public.json' to 'database.json'
    - I believe it makes sense to have a more generic name for the default message database within the final package. It would also be possible to have it still be 'messages_public.json' within the repo, but 'database.json' within the built package.
2. Type hints are provided for the `bindings.pyd` module. 
    - This module contains all functions and types which are not directly dependent on the information of the default `database.json` file.
    - Hints are provided as stub files and generated via the [CMake nanobind stub interface](https://nanobind.readthedocs.io/en/latest/typing.html#cmake-interface). 
        - This required `bindings.pyd` to be importable without being in the `novatel_edie` package. The `MessageDbSingleton::get` method had to be updated to make this possible, since it now needs to account for the possibility that no `database.json` is available.
        - The method for finding the `database.json` file was also updated to avoid importing `novatel_edie` since this could led to errors when importing `bindings.pyd` directly.
        - Since the underlying stub generation mechanism relies on importing the module, it is only possible do to on architectures which match with the built `bindings.pyd` file. This means I had to disable a couple build jobs such as `MacOS - x86_64` which builds an `x86_64` binary but does not run on an `x86_64` image.
 3. Type hints are provided for the `enums` and `messages` modules:
    - These modules may differ depending on the `database.json` file that is used.
    - Hints are provided via `enums.pyi` and `messages.pyi` stub files.
        - I noticed that many of the `Enum` definitions within `enums.pyi` where invalid in Python, as they had invalid enumerator names. For example ones that started with numbers or contained `.` characters. I updated both the stub generation logic and nanobind python enum type creation logic to convert names to valid ones.
    - The [python/novatel_edie_customizer/novatel_edie_customizer/stubgen.py](https://github.com/novatel/novatel_edie/compare/dev-riley-kinahan-python-object-rep...dev-type-hinting-support?expand=1#diff-2c9c7f2123c5ee71129f151e491f32ca8f9abf02d1d7864149ff287afce23f0d) script can be used to generate correct stubs for a given database. It is incorporated into the CMake build process as a custom target.
    - [python/novatel_edie_customizer/novatel_edie_customizer/stubgen.py](https://github.com/novatel/novatel_edie/compare/dev-riley-kinahan-python-object-rep...dev-type-hinting-support?expand=1#diff-2c9c7f2123c5ee71129f151e491f32ca8f9abf02d1d7864149ff287afce23f0d) is part of a python module called `novatel_edie_customizer` with CLI tools for generating stubs for a database, or replacing an existing `novatel_edie` installation with one configured with a different default database.
        - This module is intended as a proof of concept for how we can address the need for different databases (e.g. an extended database with private messages for internal use). If others like the idea we can split it off into its own standalone tool, otherwise we can scrap it and stick with the script.


python/bindings/message_database.cpp
```diff
@@ -183,26 +184,48 @@ void init_common_message_database(nb::module_& m)
...
     UpdatePythonEnums();
+    UpdateMessageTypes();
+}
+
+void cleanString(std::string& str)
+{
+    // Remove special characters from the string to make it a valid python attribute name
+    for (char& c : str)
+    {
+        if (c == '-' || c == '+' || c == '.' || c == '/' || c == '(' || c == ')') { c = '_'; }
```
valgur, line 226:
```suggestion
        if (!isalnum(c)) { c = '_'; }
```
I think it's simpler and safer to convert everything non-alphanumeric to an underscore.

pyproject.toml
```diff
@@ -34,10 +35,10 @@ build-backend = "scikit_build_core.build"
 
 [tool.scikit-build]
 cmake.minimum-version = "3.24"
-cmake.targets = ["python_bindings"]
+cmake.targets = ["python_bindings", "bindings_stub", "dynamic_stubs"]
 install.components = ["python"]
 wheel.license-files = ["LICENSE"]
-build-dir = "build/{wheel_tag}"
+build-dir = "out/build/{wheel_tag}"
```
valgur, line 41: Add `out/` to .gitignore as well then.

python/bindings/json_db_reader.cpp
```diff
@@ -8,31 +8,40 @@ namespace nb = nanobind;
...
     static PyMessageDatabase::Ptr json_db = nullptr;
-    if (!json_db) { json_db = std::make_shared<PyMessageDatabase>(*JsonDbReader::LoadFile(default_json_db_path())); }
+
+    // If the database has already been loaded, return it
+    if (json_db) { return json_db; }
+
+    // Import necessary modules for locating the default database
+    nb::object builtins = nb::module_::import_("builtins");
+    nb::object os_path = nb::module_::import_("os.path");
+    nb::object import_lib_util = nb::module_::import_("importlib.util");
```
valgur, line 21: I don't mind the change, but what was broken with using either `importlib_resources` or `importlib.resources`?

python/CMakeLists.txt
```diff
@@ -54,6 +54,20 @@ target_compile_definitions(python_bindings PRIVATE
...
+    MODULE bindings
+    OUTPUT stubs/bindings.pyi
+    PYTHON_PATH $<TARGET_FILE_DIR:python_bindings>
+    DEPENDS bindings
+)
+
+add_custom_target(
+    dynamic_stubs ALL
+    COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/python/novatel_edie_customizer/novatel_edie_customizer/stubgen.py ${CMAKE_SOURCE_DIR}/database/database.json ${CMAKE_CURRENT_BINARY_DIR}/stubs
+    COMMENT "Generating dynamic stubs"
```
valgur, line 68:
```suggestion
    COMMENT "Generating dynamic stubs"
    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/novatel_edie_customizer/novatel_edie_customizer/stubgen.py
```
Re-generate the stubs automatically when `stubgen.py` is modified. Makes development more convenient.

python/bindings/message_decoder.cpp
```diff
@@ -81,22 +100,20 @@ nb::object convert_field(const FieldContainer& field, const PyMessageDatabase::C
     }
 }
 
-PyDecodedMessage::PyDecodedMessage(std::vector<FieldContainer> message_, const oem::MetaDataStruct& meta_, PyMessageDatabase::ConstPtr parent_db_)
-    : fields(std::move(message_)), message_id(meta_.usMessageId), message_crc(meta_.uiMessageCrc), message_name(meta_.acMessageName), time(meta_),
-      measurement_source(meta_.eMeasurementSource), constellation(meta_.constellation), parent_db_(std::move(parent_db_))
-{
-}
+PyMessageBody::PyMessageBody(std::vector<FieldContainer> message_, PyMessageDatabase::ConstPtr parent_db_, std::string name_)
+    : fields(std::move(message_)), parent_db_(std::move(parent_db_)), name(std::move(name_))
```
valgur, line 104:
```suggestion
    : name(std::move(name_)), fields(std::move(message_)), parent_db_(std::move(parent_db_))
```
Fixes a compiler warning for fields being initialized out of order.

valgur:
Both this and #92 look very good to me, thanks!
The only thing I would change is skipping the PyMessage subclasses, for which I created a separate PR.

python/bindings/json_db_reader.cpp
riley-kinahan: If you import `bin/bindings.pyd` whilst `novatel_edie` is installed within your virtual enviroment, the old code would import `novatel_edie` in order to get information about its file location. This meant that `novatel_edie.bindings.pyd` would also get imported. The double import led to a crash.

pyproject.toml
riley-kinahan: Added some additional CMake and visual studio build ignores as well

riley-kinahan:
@valgur I'm just going to go ahead and merge this to #92 to consolidate the review. The changes I proposed should be integrated now so let me know if you have any feedback over there.

PR #97, opened 2025-02-06
Add type hinting support - drop PyMessage subclasses

valgur:
@riley-kinahan Some suggested changes to #96. Found it easier to show the changes than to add a review comment.

I think the custom PyMessage sub-class for each message type in the database is an overkill and adding a `_Body` to the message body classes adds too much noise, especially if it's included in the field names as well. This PR drops the PyMessage sub-classes and the `_Body` suffix.

I also tweaked the repr for PyMessage.

After these changes:
```python
>>> message
<Message GLOEPHEMERIS(sloto=51, freqo=0, sat_type=1, false_iod=80, ephem_week=2168, ephem_time=161118000, time_offset=10782, nt=573, GLOEPHEMERIS_reserved=0, GLOEPHEMERIS_reserved_9=0, issue=95, broadcast_health=0, pos_x=-23917966.796875, pos_y=4816388.18359375, pos_z=7425851.07421875, vel_x=-1006.2713623046875, vel_y=183.21990966796875, vel_z=-3369.5755004882812, ls_acc_x=1.862645149230957e-06, ls_acc_y=-9.313225746154785e-07, ls_acc_z=-0.0, tau=-6.69313594698906e-05, delta_tau=5.587935448e-09, gamma=0.0, tk=84600, p=3, ft=2, age=0, flags=13)>
>>> type(message)
novatel_edie.bindings.Message
>>> type(message.body)
importlib._bootstrap.GLOEPHEMERIS
>>> from novatel_edie.messages import *
... isinstance(message.body, GLOEPHEMERIS)
True
```

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -64,8 +64,11 @@ nanobind_add_stub(
 
 add_custom_target(
     dynamic_stubs ALL
-    COMMAND ${Python_EXECUTABLE} ${CMAKE_SOURCE_DIR}/python/novatel_edie_customizer/novatel_edie_customizer/stubgen.py ${CMAKE_SOURCE_DIR}/database/database.json ${CMAKE_CURRENT_BINARY_DIR}/stubs
+    COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/novatel_edie_customizer/novatel_edie_customizer/stubgen.py ${CMAKE_SOURCE_DIR}/database/database.json ${CMAKE_CURRENT_BINARY_DIR}/stubs
     COMMENT "Generating dynamic stubs"
+    DEPENDS
+        ${CMAKE_CURRENT_SOURCE_DIR}/novatel_edie_customizer/novatel_edie_customizer/stubgen.py
+        ${CMAKE_SOURCE_DIR}/database/database.json
 )
 
 if(DEFINED SKBUILD_PROJECT_NAME)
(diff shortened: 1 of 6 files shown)


riley-kinahan:
The main reason I prefer the pattern of `isinstance(msg, BESTPOS)` over `message.name == BESTPOS` is that can automatically inform IDEs of additional type info.

For example:
image
vs 
image

Of course there are other ways of manually typehinting it `body: BESTPOS = message.body` but I think making instance checks  simple and presenting them in example code is the best way to make sure users actually take advantage of the feature.


As for the `to_dict` method I can see the case for either option as the default, but I think adding an optional `include_header` parameter would let us have the best of both worlds.

valgur:
Ok, I agree, the type hinting support via `isinstance()` can be very useful, indeed. `include_header` for `to_dict()` sounds like a good compromise as well.

rldoris:
@riley-kinahan This seems to be a good compromise. I'm a huge fan of being able to assert the type and IDEs picking up the fields in the log. Huge win for usability that I hadn't been able to solve.

I think defaulting the header to true is the better option as I think more often than not people will want the log time and other fields in the header, but we can always tweak that later if needed.

riley-kinahan:
@valgur sounds like we're in agreement. How do you want to handle these updates?

The changes in this PR definitely bring us closer to the agreed upon solution. We can merge this into #96 now and I can make further updates there if that works for you.

valgur:
Sounds good. Merge this one and I'll review any further changes in the main PR.

PR #99, opened 2025-02-11
Fix profile info decoding

riley-kinahan:
`PROFILEINFO` binary log representation was being decoded incorrectly due to lack of byte-alignment between field array elements. I fixed the logic and added a unit test for the case.

I also updated the unit tests to get the database file based on their relative file location within the repo, rather than requiring it as an argument. This means no special debug config is needed and that tests can now be run via the Visual Studio test explorer.


src/decoders/common/test/main.cpp
```diff
@@ -36,9 +36,11 @@ int main(int argc, char** argv)
     testing::InitGoogleTest(&argc, argv);
     Logger::InitLogger();
 
-    if (argc != 2) { throw std::invalid_argument("1 argument required.\nUsage: <project root>"); }
+    std::filesystem::path pathSourceFile = __FILE__;
+    std::filesystem::path pathRepoDir = pathSourceFile.parent_path().parent_path().parent_path().parent_path().parent_path();
```
jonathanmcdermid, line 40: I think the issue with this method is that the relative path is dependent on where the executable is located with respect to the root directory, which can change with some build configurations. @rldoris shot this idea down when I proposed it iirc.

riley-kinahan: I _think_ this shouldn't be an issue. `__FILE__` is a preprocessor directive which gets substituted with the source file path rather than the executable path. Pretty sure it should work as long as the build is done on the same machine as the test is run.

image

jonathanmcdermid: Oh I see, youre using the filepath rather than the executable path. I like this idea.

jonathanmcdermid (approved):
looks good!

PR #101, opened 2025-02-21
High level access redesign

riley-kinahan:
Core C++ library changes:
- Added a copy constructor to `MetaDataStruct`/`MetaDataBase`
- Made `Encoder`/`EncoderBase` capable of being initialized with either a const raw or const shared pointer to a `MessageDatabase`. This is necessary to allow a `PyMessageDatabase` to own a pointer to an `Encoder` and allow messages to decode themselves. 
- Made `Encoder.Encode` only take a `HEADER_FORMAT` value rather than a `MetaDataStruct` reference. If anything this should be cheaper and it clarifies what data is used by the `Encode` process.
- Split `Parser.Read` into `Parser.Read` and `Parser.ReadIntermediate` to allow more transparent access to the intermediate message formats that the python API needs access to.
- Split `FileParser.Read` into `FileParser.Read` and `FileParser.ReadIntermediate` mirroring what I did with the `Parser` class. Used `HandleRead` function to avoid code duplication between them.

Python Interface changes:
- Allow `Parser` /`FileParser` to iterate through `Message` objects.
- Allow `Parser` /`FileParser` to iterate through convert messages via `convert()`.
- Allow `Parser` /`FileParser` to iterate through converted messages via `iter_convert()`.
- Made `Framer` iterable and have it return frame(`bytes`) and metadata(`MetaData`). Abstracted passing of `MetaData` into the framer.
- Combined `HeaderDecoder` and `MessageDecoder` into a single class, with `decode`, `decode_header`, and `decode_message` methods.
- Made passing `MetaData` to decode functions optional. Still an option to get additional data out or for processing more efferently. 
- Removed `JsonDbReader` and moved functionality to `MessageDatabase` constructor.
- Added the ability for a message object to encode itself e.g. `msg.to_ascii()`
- No longer return `STATUS` values anywhere, instead raise appropriate custom exception.


riley-kinahan:
@valgur I still have work to do on this one: `Framer`, `RxconfigHandler` and `Commander` are not updated yet, I also need to fix the unit tests.

I've opened this up as a draft in case you want to look into what I've done so far. I added a brief summary to the PR describing the changes but I'll also fixing up the examples and adding some better documentation soon.

python/bindings/py_message_objects.cpp
```diff
@@ -0,0 +1,409 @@
...
+}
+
+#pragma endregion
+
+#pragma region PyMessageMethods
+
+PyMessageData PyCompleteMessage::to_ascii() { 
+    PyEncoder encoder = PyEncoder(this->parent_db_);
+    return encoder.PyEncode(*this, ENCODE_FORMAT::ASCII);
+}
```
riley-kinahan, line 224: Will try to improve the implementation of this to be less inefficient and fix remaining unit tests. After that this PR is ready for full review.

python/bindings/py_database.hpp
```diff
@@ -56,7 +64,12 @@ class PyMessageDatabase final : public MessageDatabase
     std::unordered_map<std::string, nb::object> enums_by_id{};
     std::unordered_map<std::string, nb::object> enums_by_name{};
 
+    std::shared_ptr<oem::Encoder> encoder;  // This must always be a PyEncoder
```
riley-kinahan, line 67: I would like to make this a pointer to an actual `PyEncoder` type however I wasn't able to figure out a way to do so while avoiding circular dependency. I'd appreciate if anyone has suggestions on improving the  design.

riley-kinahan:
Coding work for this PR is done pending feedback, unit tests are also updated to pass. I may continue to work on documentation, examples, and adding additional context to the PR itself.

python/readme.md
```diff
@@ -0,0 +1,196 @@
...
+    elif isinstance(msg, IncompleteMessage):
+        print(msg.header)
+    elif isinstance(msg, BESTPOS):
+        print(msg.latitude)
+```
+Under certain circumstances parsing may encounter an error,
+if you wish to handle this it is recommended to iterate through 
+the `FileParser`/`Parser` object manually via the `read()` method.
+```python
+buffer_size = MAX_BINARY_MESSAGE_LENGTH
```
valgur, line 124:
```suggestion
```
This line does not appear relevant to the example?

python/readme.md
```diff
@@ -0,0 +1,196 @@
...
+while True:
+    try:
+        msg = parser.read()
+    except Exception as e:
+        logging.exception('An exception occurred', e)
```
valgur, line 129: I think an example of a realistic exception raised by EDIE would be useful here.

python/readme.md
```diff
@@ -0,0 +1,196 @@
...
+for msg in parser:
+    ...
+try:
+    last_msg = parser.read(decode_incomplete_abbreviated = True)
+except Exception:
+    pass
+```
+
+Direct format conversion is also supported via the `convert` method:
+```
```
valgur, line 145:
```suggestion
```python
```

python/readme.md
```diff
@@ -0,0 +1,196 @@
...
+for ascii_msg in parser.convert(ENCODE_FORMAT.ASCII):
```
valgur, line 146: The behavior of `.convert()` method seems a bit unintuitive at first glance. How about `parser.read(convert=ENCODE_FORMAT.ASCII)` as the API instead?

python/bindings/decoder.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+class PyDecoder
+{
+  public:
+    PyMessageDatabase::Ptr database;
+    HeaderDecoder header_decoder;
+    MessageDecoder message_decoder;
+
+    PyDecoder()
+        : database(MessageDbSingleton::get()), header_decoder(HeaderDecoder(MessageDbSingleton::get())), message_decoder(MessageDbSingleton::get())
```
valgur, line 23:
```suggestion
        : PyDecoder(MessageDbSingleton::get())
```
Can simplify it a bit with a delegating constructor.

python/bindings/bindings.cpp
```diff
@@ -43,4 +48,5 @@ NB_MODULE(bindings, m)
     init_novatel_oem_messages(messages_mod);
     nb::module_ enums_mod = m.def_submodule("enums", "Enumerations used by NovAtel OEM message fields.");
     init_novatel_oem_enums(enums_mod);
+    init_decoder_tester(m);
```
valgur, line 51: I would move the internal helper classes to a separate submodule as well.

python/bindings/exceptions.cpp
```diff
@@ -0,0 +1,55 @@
...
+    case STATUS::NO_DATABASE: throw NoDatabaseException();
+    case STATUS::NO_DEFINITION: throw NoDefinitionException();
+    case STATUS::NO_DEFINITION_EMBEDDED: throw NoDefinitionEmbeddedException();
+    case STATUS::BUFFER_FULL: throw BufferFullException();
+    case STATUS::BUFFER_EMPTY: throw BufferEmptyException();
+    case STATUS::STREAM_EMPTY: throw StreamEmptyException();
+    case STATUS::UNSUPPORTED: throw UnsupportedException();
+    case STATUS::MALFORMED_INPUT: throw MalformedInputException();
+    case STATUS::DECOMPRESSION_FAILURE: throw DecompressionFailureException();
+    default: throw std::runtime_error("Unknown STATUS value");
```
valgur, line 32:
```suggestion
    default: throw std::runtime_error("Unknown STATUS value: " + std::to_string(status));
```

python/bindings/exceptions.cpp
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+void init_novatel_exceptions(nb::module_& m) {
+    nb::exception<FailureException>(m, "FailureException"); // NOLINT
```
valgur, line 37: I think it would be great if you could add a base exception class for all NovAtel EDIE exceptions and derive the rest from it. Would make it easier to catch relevant exceptions only.

python/bindings/file_parser.cpp
```diff
@@ -1,67 +1,99 @@
...
+    header.format = metadata.eFormat;
+
+    nb::object pyinst;
+    switch (status)
+    {
+    case STATUS::SUCCESS: pyinst = create_message_instance(header, message_fields, metadata, parent_db); break;
+    case STATUS::NO_DEFINITION:
+        pyinst = create_unknown_message_instance(nb::bytes(message_data.pucMessageBody, message_data.uiMessageBodyLength), header, parent_db);
+        break;
+    case STATUS::UNKNOWN: pyinst = nb::bytes(message_data.pucMessage, message_data.uiMessageLength); break;
```
valgur, line 32: You can probably return directly instead of assigning to a temporary var?

python/bindings/framer.cpp
```diff
@@ -21,23 +22,32 @@ void init_novatel_framer(nb::module_& m)
...
-                {
-                    oem::MetaDataStruct metadata;
-                    STATUS status = framer.GetFrame(reinterpret_cast<uint8_t*>(buffer.data()), buffer_size, metadata);
-                    return nb::make_tuple(status, nb::bytes(buffer.data(), metadata.uiLength), metadata);
+                case STATUS::UNKNOWN:
+                    frame_vals = nb::make_tuple(nb::bytes(buffer.data(), metadata.uiLength), oem::MetaDataStruct(metadata));
+                    break;
+                case STATUS::SUCCESS:
+                    frame_vals = nb::make_tuple(nb::bytes(buffer.data(), metadata.uiLength), oem::MetaDataStruct(metadata));
+                    break;
```
valgur, line 37: Can return directly?

python/readme.md
riley-kinahan: I think `parser.read(convert=ENCODE_FORMAT.ASCII)` is a nice alternative interface for getting a single encoded message, but I still like the idea of being able to iterate directly on converted messages.

If your goal is to get all messages within a file in ascii format

```python
ascii_msgs = [msg for msg in parser.convert(ENCODE_FORMAT.ASCII)]
``` 

is simpler than

```python
ascii_msgs = []
while True:
    try:
         msgs.append(parser.read(ENCODE_FORMAT.ASCII))
    except StopIteration:
         break
```

Do you think renaming the method might make it clearer?

python/bindings/exceptions.cpp
riley-kinahan: I agree but as far as I can tell nanobind only provides native support for sub classing built-in python exceptions: https://nanobind.readthedocs.io/en/latest/exceptions.html

python/readme.md
riley-kinahan: I like keeping `read()` as an option for directly handling each parse action. I think that paradigm will be preferred by users if they are doing something like attempting to decode a message every time they receive data through a COM port.

valgur: How about a separate `.read_single()` for that case?

I would personally not enjoy having `while True: try: ... catch StopIteration:` as the default usage pattern for the library.

riley-kinahan: As it stands right now the default usage pattern is:
```python
for msg in parser:
    msg = msg.to_ascii()    
```

The `convert` method is offered as a way to more directly and efficiently iterate over converted messages
```python
for msg in parser.convert(encode_format):
    ...
```

Then the `read` method gives you more fine grained control over each action
```python
while True:
    try:
        msg = parser.read()
    ...
```

This is my understanding of your proposal

Default usage pattern:
```python
for msg in parser.read():
    ...
```

Direct conversion:
```python
for msg in parser.read(encode_format):
    ...
```

One-by-one conversion:
```python
while True:
    try:
        msg = parser.read_single()
    ...
```

The main concern I have with this approach is that iterating through data and decoding it is meaningfully different than iterating through to decode and then re-encode it. `parser.read()` and `parser.read(ENCODE_FORMAT.ASCII)` look like their return values should be of the same kind, but the logic to handle either one would have to differ significantly.

How would you feel about something like `read()`, `read_single()`, `convert()`, and `convert_single()` to make the interface for reading and converting more clear but maintain a level of separation between the operations?

riley-kinahan:
FYI @jonathanmcdermid I removed the python `load_json_db_file` and `parse_json_db` you introduced in #102. The same functionality is instead within a constructor and static method of the `MessageDatabase` class.

python/bindings/exceptions.cpp
riley-kinahan: Turns out it is supported! I just did it the wrong way initially. 

There is now an `NovatelEdieException` base class.

python/readme.md
valgur: I think the current pattern of `for msg in parser:` providing an iterator and `parser.read()` providing a more fine-grained one-by-one access is very reasonable. A separate `read()` and `read_single()` would be redundant.
My only concern is that the `.convert()` method name sounds like it would probably be used as `converted_msg = parser.convert(input_msg, format)` and not like something that would advance and modify the state of the parser. It's a bit cumbersome, but something like `.read_and_convert()` would be better in that regard, imo.

riley-kinahan: I agree its not clear, ended up going with `convert()` to convert a single message and  `iter_convert()` to get an iterator.

riley-kinahan:
@valgur Will be having a discussion with the other devs about these changes tomorrow. PR should get merged fairly soon barring additional concerns being raised.

python/bindings/bindings.cpp
valgur: I was thinking something like 
```suggestion
    nb::module_ internal_mod = m.def_submodule("internal", "Internal helper functions.");
    init_decoder_tester(internal_mod);
```

python/readme.md
valgur: That's perfect. Thanks!

include/novatel_edie/decoders/common/encoder.hpp
```diff
@@ -649,12 +651,18 @@ template <typename Derived> class EncoderBase
     //
     //! \param[in] pclMessageDb_ A pointer to a MessageDatabase object. Defaults to nullptr.
     //----------------------------------------------------------------------------
-    EncoderBase(MessageDatabase::Ptr pclMessageDb_ = nullptr)
+    EncoderBase(const MessageDatabase* pclMessageDb_ = nullptr)
```
rldoris, line 654: This looks like a regression.

python/bindings/bindings.cpp
riley-kinahan: Ah I see, good suggestion!

python/bindings/file_parser.cpp
```diff
@@ -1,67 +1,123 @@
...
-            "set_stream", [](oem::FileParser& self, nb::object stream) { return self.SetStream(std::make_shared<pystream::istream>(stream, 0)); },
+            "set_stream", [](oem::PyFileParser& self, nb::object stream) { return self.SetStream(std::make_shared<pystream::istream>(stream, 0)); },
             "input_stream"_a)
-        .def("read",
-             [](oem::FileParser& self) {
-                 MessageDataStruct message_data;
-                 oem::MetaDataStruct meta_data;
-                 STATUS status = self.Read(message_data, meta_data);
-                 return std::make_tuple(status, oem::PyMessageData(message_data), meta_data);
+        .def("read", &oem::PyFileParser::PyRead)
```
rldoris, line 99: Do you have any idea on how much overhead this new solution has? Have you done any performance testing?

include/novatel_edie/decoders/common/encoder.hpp
riley-kinahan: Yes this does make it easier to create memory unsafe code on the C++ side but it is required for messages to decode themselves. C++ code should always instantiate an `Encoder` with a shared pointer if the `Encoder` is responsible for the lifetime of its `MessageDatabase`, this is the case right now but could lead to errors down the road.

A `PyMessageDatabase`  (which `Message` objects hold a reference to) stores a shared pointer to an `Encoder` object in order to allow decoding. However, an `Encoder` also must own a reference to a `MessageDatabase` in order to do its job.  If we don't allow this reference to be a weak one it creates a cycle and memory that will never be freed.

python/bindings/parser.cpp
```diff
@@ -1,63 +1,107 @@
...
 #include "message_db_singleton.hpp"
+#include "parser.hpp"
 #include "py_message_data.hpp"
+#include "py_message_objects.hpp"
 
 namespace nb = nanobind;
 using namespace nb::literals;
 using namespace novatel::edie;
 
+nb::object oem::PyParser::PyRead(bool decode_incomplete)
```
rldoris, line 14: This is very similar to the PyRead in fileparser. Could they be combined or use a common function to reduce code duplication?

python/bindings/file_parser.cpp
riley-kinahan: Top level file parsing is marginally slower than our old DLL-based python package. 95.2s vs 90.5s for a ~4GB file. I haven't dug much deeper than that.

python/bindings/parser.cpp
riley-kinahan: I think its hard to combine any of the code until after the `ReadIntermediate` function call, without making the code a lot more complex. 

We could create a function called `HandlePythonReadStatus` just to handle the switch statement at the end. Otherwise I think we are better off living with the duplication.

python/examples/converter_fileparser.py
```diff
@@ -45,6 +45,7 @@ def _configure_logging(logger):
 
 
 def main():
+    pass
```
rldoris, line 48: Remove

python/examples/converter_fileparser.py
```diff
@@ -66,36 +67,36 @@ def main():
...
-
-            converted_logs_stream.write(message_data.message)
-            logger.info(f"Encoded: ({len(message_data.message)}) {message_data.message}")
-            complete_messages += 1
-
-            if timeit.default_timer() - loop > 1:
-                counter += 1
-                logger.info(f"{file_parser.percent_read}% {complete_messages / counter} logs/s")
-                loop = timeit.default_timer()
+    # _configure_logging(file_parser.filter.logger)
```
rldoris, line 76: Remove

python/bindings/file_parser.cpp
rldoris: Very reasonable

riley-kinahan:
Made the following adjustments following interface design discussion:
- Renamed and reorganized data types to improve clarity:
    - `CompleteMessage` -> `Message`.
    - `IncompleteMessage` -> `UnknownMessage`.
    - `Message` -> `MessageBase`.
    - Created `UnknownBytes` type to be returned by `FileParser`/`Parser` instead of regular `bytes`.
    - Created `EdieData` superclass to unify return types. 

New type hierarchy:
```mermaid
classDiagram
    class EdieData {
    }
    
    class UnknownBytes {
        +data: bytes
    }

    class MessageBase {
        +header: Header
   }

    class UnknownMessage {
        +payload: bytes
    }
    
    class Message {
    }
    
    class BESTPOS {
        + lat: float
        + long: float
    }

    class RANGE {
        + obs: list
    }


    EdieData<|-- UnknownBytes
    EdieData <|-- Field 
    Field <|-- MessageBase
    MessageBase<|-- UnknownMessage
    MessageBase<|-- Message
    Message <|-- BESTPOS
    Message <|-- RANGE
```
- Removed `Encoder` object since `encode`/`to_ascii`/`to_binary` methods capture all current functionality.
- Added function signatures and docstrings to `read` and `__next__` methods of `Parser` and `FileParser`. 

With these changes I believe this pull-request is ready to be merged. Once it has, I will open another one to update logging to be managed by the `logging` module.

riley-kinahan:
@rldoris switched to the flatter heiarchy you suggested while still keeping `UnknownBytes`:

```mermaid
classDiagram
    
    class UnknownBytes {
        +data: bytes
    }


    class UnknownMessage {
        +payload: bytes
    }
    
    class Message {
    }
    
    class BESTPOS {
        + lat: float
        + long: float
    }

    class RANGE {
        + obs: list
    }

 
    Field <|-- Message
    Message <|-- BESTPOS
    Message <|-- RANGE
```

python/bindings/file_parser.cpp
```diff
@@ -1,67 +1,139 @@
...
-    nb::class_<oem::FileParser>(m, "FileParser")
-        .def("__init__", [](oem::FileParser* t) { new (t) oem::FileParser(MessageDbSingleton::get()); }) // NOLINT(*-cplusplus.NewDeleteLeaks)
-        .def(nb::init<const std::filesystem::path&>(), "json_db_path"_a)
-        .def(nb::init<const PyMessageDatabase::Ptr&>(), "message_db"_a)
-        .def("load_json_db", &oem::FileParser::LoadJsonDb, "json_db_path"_a)
-        .def_prop_ro("logger", &oem::FileParser::GetLogger)
-        .def("enable_framer_decoder_logging", &oem::FileParser::EnableFramerDecoderLogging, "level"_a = spdlog::level::debug,
+    nb::class_<oem::PyFileParser>(m, "FileParser")
+        .def(nb::init<const std::filesystem::path&>(), "file_path"_a)
+        .def(nb::init<const std::filesystem::path&, const PyMessageDatabase::Ptr&>(), "file_path"_a, "message_db"_a)
```
valgur, line 80: Since Python stream input is supported via `.set_stream()` could file-like input (e.g. a file handle or BytesIO) be to the constructor as well, perhaps? Would make it much more convenient to use. Besides, `.set_stream` is a bit useless if the constructor needs to be passed a real file path first.

If not, dropping the Python stream support altogether might be a valid option as well (relying on `Parser.write()/read()` instead), since the Python stream wrapper is something that could have many sharp edges hiding. But I would prefer keeping the support, if possible.

valgur:
I have to say that I'm incredibly happy with the improvements being made here to the API. I needed the library to analyze some data today and being able to simply do the equivalent of
```python
In [1]: import novatel_edie as ne
   ...: decoder = ne.Decoder()
   ...: decoder.decode(bytes([170, 68, 18, 28, 101, 0, 0, 64, 44, 0, 0, 0, 154, 120, 54, 9, 136, 13, 23, 13, 32, 193, 72, 2, 36, 153, 23, 64, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 48, 105, 71, 225, 70, 220,
   ...:  53, 63, 197, 123, 255, 255, 255, 255, 49, 192, 233, 7, 0, 0, 3, 18, 12, 59, 24, 221, 0, 0, 1, 0, 0, 0, 142, 3, 109, 35]))
[2025-03-18 15:38:31.630] [root] [info] Logger initialized.
Out[1]: TIME(clock_model_status=<ClockModelStatus.INVALID: 3>, offset=0.0, offset_std=0.00033356409519815205, utc_offset=-17.999999999879737, utc_year=2025, utc_month=3, utc_day=18, utc_hour=12, utc_minute=59, utc_millisecond=56600, utc_time_status=<UTCTimeStatus.VALID: 1>)
```
is so-so satisfying.

Should change the `Logger initialized.` level to `DEBUG`, though. :slightly_smiling_face:

python/bindings/file_parser.cpp
riley-kinahan: Oh this is a really good point!

The best solution might be a dedicated `StreamParser` class which can be instantiated with any `io.IoBase` object. I think that would result in the clearest interface and allow flexibility for handling any edge cases. Since adding it would be non-breaking and `Parser` can be used as a workaround, I'd prefer to leave it to a future PR and just remove the `.set_stream` method for now.

riley-kinahan:
@valgur happy to hear you're enjoying the changes!

FYI my next PR is specifically a logging rework so I'll switch the logger initialization message level there.

python/bindings/py_message_objects.hpp
```diff
@@ -0,0 +1,145 @@
...
+    uint16_t week{0};
+    double milliseconds{0.0};
+    TIME_STATUS time_status{TIME_STATUS::UNKNOWN};
+};
+
+//============================================================================
+//! \class UnknownBytes
+//! \brief A series of bytes determined to be undecodable.
+//============================================================================
+struct PyUnknownBytes
```
rldoris, line 30: Could you subclass bytes so it still acted like a bytes object? 

Chatgtp code: No idea if it works

```python

struct unknownbytes : nb::bytes {
    // Inherit constructors explicitly
    using nb::bytes::bytes;

    // Optionally, add custom constructors or methods
};```

rldoris (approved):

riley-kinahan:
Going ahead and merging this, Python API is close to where it needs to be but should still be considered unstable at this point. My current plan is to create a PR to update the logging management, and then another cleanup PR with improved docs and possibly some smaller interface tweaks.

PR #102, opened 2025-02-22
Delete JsonDbReader

jonathanmcdermid:
This pull request includes several changes aimed at refactoring the codebase to simplify the loading and parsing of JSON database files. The most important changes include removing the `JsonDbReader` class and moving its methods to the global namespace, updating various files to use the new method signatures, and improving the `MessageDatabase` class by separating the generation of mappings for messages and enumerations.

Refactoring of JSON database loading and parsing:

* [`include/novatel_edie/decoders/common/json_db_reader.hpp`](diffhunk://#diff-d7d08da9ea1afd2b67dfcdb4be36be2e6d207ae85ae9bdadca90bccab204f24eL57-R66): Removed the `JsonDbReader` class and moved its static methods (`LoadFile`, `Parse`, `AppendMessages`, and `AppendEnumerations`) to the global namespace. [[1]](diffhunk://#diff-d7d08da9ea1afd2b67dfcdb4be36be2e6d207ae85ae9bdadca90bccab204f24eL57-R66) [[2]](diffhunk://#diff-d7d08da9ea1afd2b67dfcdb4be36be2e6d207ae85ae9bdadca90bccab204f24eL82-R91)
* [`src/decoders/common/src/json_db_reader.cpp`](diffhunk://#diff-18f2f32608d8a2eba5fba4a30894192aa071dd7bbbc3f27566a88d1fd8ae38e8R29-R30): Included necessary headers and updated the implementation to reflect the changes in `json_db_reader.hpp`.

Updates to benchmark and example files:

* [`benchmarks/benchmark.cpp`](diffhunk://#diff-f67013302448def8fb420590ad44ab1b089aa9b98890d20076a6b72aa08e10abL44-R51): Replaced calls to `JsonDbReader::LoadFile` with `LoadFile`. [[1]](diffhunk://#diff-f67013302448def8fb420590ad44ab1b089aa9b98890d20076a6b72aa08e10abL44-R51) [[2]](diffhunk://#diff-f67013302448def8fb420590ad44ab1b089aa9b98890d20076a6b72aa08e10abL134-R134) [[3]](diffhunk://#diff-f67013302448def8fb420590ad44ab1b089aa9b98890d20076a6b72aa08e10abL175-R175) [[4]](diffhunk://#diff-f67013302448def8fb420590ad44ab1b089aa9b98890d20076a6b72aa08e10abL213-L224)
* Updated example files to use the new `LoadFile` method instead of `JsonDbReader::LoadFile`. [[1]](diffhunk://#diff-1d8ce26aeafea3247ef75bb398ddbc5707689339ef04ce265538f2fc50d61403L78-R78) [[2]](diffhunk://#diff-7bd312629dd8ec9b7f4a49344057960790f78b20a8ea58edcdb1bc55f7e0730bL94-R94) [[3]](diffhunk://#diff-30aa92d4573b99d5766e1899bde598295e2b1738bc02393b52a17dabe2c9c1f8L90-R90) [[4]](diffhunk://#diff-b8afddc941a0da451489dc9ae0b8e029b2f6cda048d54727cd96df8b5d648f18L88-R88) [[5]](diffhunk://#diff-bb9df76accd86d931d526ac2ae4589a22ae3b6b272f5a7b7bddc66f103027d7fL91-R97) [[6]](diffhunk://#diff-4ee1fb549e13777d8b933b86d947bc4ca907e8e990438fe567ccd08571da467eL96-R96) [[7]](diffhunk://#diff-8ff05ff52023b456fbbce45340a7de3eb957dded6546d26296af02ce8e2ecd42L89-R89)

Improvements to `MessageDatabase` class:

* [`include/novatel_edie/decoders/common/message_database.hpp`](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL447-R448): Split the `GenerateMappings` method into `GenerateEnumMappings` and `GenerateMessageMappings` for better code clarity and separation of concerns. [[1]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL447-R448) [[2]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL460-R462) [[3]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL489-R492) [[4]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL509-R513) [[5]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL525-R529) [[6]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL541-R545) [[7]](diffhunk://#diff-0fc55a5be68b0ad9d0b753bc98107e94ccc79138201988c81aee61498061297bL621-R635)

Python bindings update:

* [`python/bindings/json_db_reader.cpp`](diffhunk://#diff-43f1ce9d3aa5b94ad99f38f9e42e0ed93880e103d045ae3160ba982414261ec5L45-R71): Updated the Python bindings to reflect the removal of the `JsonDbReader` class and the migration of its methods to the global namespace.


src/decoders/common/src/json_db_reader.cpp
```diff
@@ -190,125 +192,102 @@ void ParseEnumerators(const json& j_, std::vector<EnumDataType>& vEnumerators_)
...
 
+void processMessageSubset(const json& messages, std::vector<MessageDefinition::ConstPtr>& result)
+{
+    for (const auto& it : messages) { result.emplace_back(std::make_shared<MessageDefinition>(it)); }
+}
+
+void processEnumSubset(const json& enums, std::vector<EnumDefinition::ConstPtr>& result)
+{
+    for (const auto& it : enums) { result.emplace_back(std::make_shared<EnumDefinition>(it)); }
+}
```
riley-kinahan, line 203: Were these functions leftover from some experimentation? I'm not sure I understand their role in the current implementation. 

Otherwise these changes look good to me, I shouldn't be much work to adapt #101 to them.

jonathanmcdermid: good eye, those were leftover from when I was looking into launching multiple threads for the subarrays but the message files ended up being too small for it to be worth it

valgur:
I would make the function names more specific and clear:
- `LoadFile` -> `LoadDbFile` or even `LoadJsonDbFile`,
- `Parse` -> `ParseDb` or even `ParseJsonDbFile`.

I would personally leave the door open to more efficient serialization formats for the DB, which the JSON DB could be transformed into at build time, for example. Even a simple GZip compression would not hurt (7 MB -> 170 kB).

Both of the `Append` functions could also be made less specific to the JSON format by replacing them with `MessageDatabase::Append` for something like
```cpp
pclMyJsonDb->Append(LoadDbFile(...));
```

What do you think?

jonathanmcdermid:
@valgur I've tried to apply your suggestions, let me know what you think. I also just deleted the appendmessage and enum functions from json_db_reader because I think you can do the same thing using the appendJsonDb methods for MessageDatabase. If for some reason we need these functions the way they were, it seems to me it would be better to have them in json_db_reader otherwise we have to add json reading logic to MessageDatabase and end up with some circular dependencies.

jonathanmcdermid:
@valgur I tried to apply the changes you suggested, take a loot and let me know what you think. I also just deleted the appendMessages/enums functions from json db reader because the same functionality is already possible with existing methods in message database. If we want to be able to do it directly with a filepath, then we have to add json reading logic to messageDatabase and end up with some weird circular dependencies. this is how I'm accomplishing the same thing.

`    clJsonDb->AppendJsonDbMessages(LoadJsonDbFile(sAppendMsg)->MessageDefinitions());
`

valgur (approved):
Thanks!

rldoris (approved):

PR #103, opened 2025-02-26
rule 0/3/5

jonathanmcdermid:
enforce rule of 0/3/5, use compiler generated methods when appropriate. move basic function definitions to headers


include/novatel_edie/decoders/common/message_database.hpp
rldoris, line 1: Are you removing all of the copy & equality functions because the compiler generated version is sufficient? You're sure there aren't using objects on the heap?

include/novatel_edie/decoders/common/common.hpp
```diff
@@ -266,19 +266,9 @@ class MetaDataBase
...
-    char acMessageName[uiMessageNameMax + 1]{};
-
-    [[nodiscard]] std::string MessageName() const { return {acMessageName}; }
-
-    void MessageName(std::string_view strMessageName_)
-    {
-        memcpy(acMessageName, strMessageName_.data(), strMessageName_.length());
-        acMessageName[strMessageName_.length()] = '\0';
-    }
+    std::string acMessageName;
```
rldoris, line 269: This should be sMessageName. No longer an array of chars.

include/novatel_edie/decoders/oem/rangecmp/common.hpp
```diff
@@ -414,8 +414,6 @@ struct RangeData
     float fCNo{0};
     float fLockTime{0};
     uint32_t uiChannelTrackingStatus{0};
-
-    RangeData() = default;
```
rldoris, line 418: What's the value of this? Just to remove unnecessary code or clarity? I'm reasonably ok either way, but would like to know your thoughts.

If we have default for one of the 5 should we include all 5 (either with implementations or default)?

python/bindings/nexcept.cpp
rldoris, line 1: @riley-kinahan and @valgur are discussing exception handling in #101 
Take a look and see if we should hold off on changing nexcept code till they have figured out exception handling on their end.

include/novatel_edie/decoders/common/message_database.hpp
jonathanmcdermid: Yep those structs will have compiler generated constructors that are sufficient

include/novatel_edie/decoders/oem/rangecmp/common.hpp
jonathanmcdermid: My understanding is that for rule of 3 and 5, if you have to create a custom function for one, then you also probably need a custom implementation for the others. Conversely, if one of them doesnt require a custom implementation, then they probably all dont need one.

python/bindings/nexcept.cpp
riley-kinahan: Currently I'm not using nexcept for raising anything on the Python side. I am trying to see if there's a way to get exception subtypes working with nanobind, but I don't think I'll need this code for that.

rldoris (approved):

PR #118, opened 2025-03-12
perf optimizations

jonathanmcdermid:
I did some messing around with perf to see what the hotspots in our code are. The main changes are below.

- remove snprintf
- use simdjson over nlohmann
- avoid copying strings
- call the more efficient method of copying to a buffer when possible
- use move semantics to reduce ref counting for definitions
- use maps for O(1) lookups in enum definitions
- use from_chars instead of c functions

Creating less copies of the data and using doing stuff in place with string views is also a good idea for future improvement. If we could change the CRC polynomial, we could make binary encoding about 3x faster using hardware intrinsics.

Before:
Screenshot from 2025-03-18 18-09-06
After:
Screenshot from 2025-04-11 17-41-42


rldoris:
Do you have any stats about the improvement in encoding speed that you could provide?

include/novatel_edie/decoders/common/encoder.hpp
```diff
@@ -471,22 +477,23 @@ template <typename Derived> class EncoderBase
...
+        case DATA_TYPE::CHAR: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<int8_t>(fc_.fieldValue));
+        case DATA_TYPE::USHORT: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<uint16_t>(fc_.fieldValue));
+        case DATA_TYPE::SHORT: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<int16_t>(fc_.fieldValue));
+        case DATA_TYPE::UINT: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<uint32_t>(fc_.fieldValue));
+        case DATA_TYPE::INT: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<int32_t>(fc_.fieldValue));
+        case DATA_TYPE::ULONG: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<uint32_t>(fc_.fieldValue));
+        case DATA_TYPE::LONG: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<int32_t>(fc_.fieldValue));
+        case DATA_TYPE::ULONGLONG: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<uint64_t>(fc_.fieldValue));
+        case DATA_TYPE::LONGLONG: return PrintToBuffer(ppcOutBuf_, uiBytesLeft_, "{}", std::get<int64_t>(fc_.fieldValue));
```
rldoris, line 496: If 99% of the time our format string is `{}` should we just default it to that so we don't have to enter it every time? Is is that obfuscating it's behavior?

jonathanmcdermid:
> Do you have any stats about the improvement in encoding speed that you could provide?

Before:
-------------------------------------------------------------------
Benchmark                         Time             CPU              Iterations
-------------------------------------------------------------------
EncodeAsciiLog                 2667 ns         2666 ns       261495       logs_per_second=375.044k/s
EncodeAbbrevAsciiLog    2367 ns         2366 ns       287237       logs_per_second=422.581k/s
EncodeJsonLog                 2689 ns         2689 ns       260123       logs_per_second=371.882k/s

After:
-------------------------------------------------------------------
Benchmark                         Time             CPU              Iterations
-------------------------------------------------------------------
EncodeAsciiLog                 1937 ns         1936 ns       362343        logs_per_second=516.535k/s
EncodeAbbrevAsciiLog     1618 ns        1617 ns       436032        logs_per_second=618.287k/s
EncodeJsonLog                  1941 ns         1941 ns       362965       logs_per_second=515.287k/s

25-35% drop in CPU time for ASCII encoding benchmarks, resulting in 35-50% increase in logs/second

include/novatel_edie/decoders/common/encoder.hpp
jonathanmcdermid: how would that work? we still need different behaviour for each case based on the std::get template argument.

jonathanmcdermid: I agree that sounds like a good idea but I dont see how it could be implemented. the default arg needs to be at the back of the parameter list and it would be interpreted as part of the variadic if we do that.

rldoris: Just a pattern I noticed. If it's not to possible or not worth the effort then I won't hold up the change.

.github/workflows/build_python.yml
```diff
@@ -7,7 +7,7 @@ on:
       python_versions:
         description: 'Which versions to build wheels for'
         required: false
-        default: cp39* cp313*
+        default: cp313*
```
riley-kinahan, line 10: I'd like to keep compatibility back to 3.9 if possible since it still has [official support](https://www.python.org/downloads/#:~:text=3.9-,security,-2020%2D10%2D05). I know you were running into some unit test failures but I think we should at minimum keep a stable 3.11 build since its the most common version around here.

conanfile.py
```diff
@@ -46,6 +46,7 @@ def layout(self):
         cmake_layout(self)
 
     def requirements(self):
+        self.requires("simdjson/3.10.1")
         self.requires("nlohmann_json/[>=3.11 <3.12]", transitive_headers=True)
```
riley-kinahan, line 50: Have we fully dropped `nlohmann_json` in this PR? If so it can be removed as a dependency.

src/decoders/oem/src/header_decoder.cpp
```diff
@@ -86,20 +86,63 @@ template <ASCII_HEADER eField> bool HeaderDecoder::DecodeAsciiHeaderField(Interm
...
+        if (result.ec != std::errc()) { throw std::runtime_error("Failed to parse SEQUENCE"); }
+        stInterHeader_.usSequence = usSequence;
         break;
-    case ASCII_HEADER::SEQUENCE: stInterHeader_.usSequence = static_cast<uint16_t>(strtoul(*ppcLogBuf_, nullptr, 10)); break;
-    case ASCII_HEADER::IDLE_TIME: stInterHeader_.ucIdleTime = static_cast<uint8_t>(std::lround(2.0F * strtof(*ppcLogBuf_, nullptr))); break;
+    }
+    case ASCII_HEADER::IDLE_TIME: {
+        float fIdleTime = 0.0F;
+        auto result = std::from_chars(*ppcLogBuf_, *ppcLogBuf_ + ullTokenLength, fIdleTime);
+        if (result.ec != std::errc()) { throw std::runtime_error("Failed to parse IDLE_TIME"); }
```
riley-kinahan, line 101: I'd suggest logging the failure reason and returning `false` here.

I think we should start avoiding raising exceptions unless it reflects and unrecoverable state, since the the parser and file parser only handle return values and not exceptions.

.github/workflows/build_python.yml
rldoris: I agree that 3.9 minimum seems reasonable. What's the impact of this?

rldoris:
Can we get some updated benchmark numbers?

.github/workflows/build_cpp.yml
```diff
@@ -52,13 +52,13 @@ jobs:
         shell: bash
         run: |
           if [ "${{ env.COMPILER }}" == "msvc" ]; then
-            cmake -S . -B out/build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DWARNINGS_AS_ERRORS=ON
+            cmake -S . -B out/build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -DBUILD_BENCHMARKS=ON -DWARNINGS_AS_ERRORS=ON
           elif [ "${{ env.COMPILER }}" == "clang-cl" ]; then
-            cmake -S . -B out/build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -T ClangCL #TODO: -DWARNINGS_AS_ERRORS=ON
+            cmake -S . -B out/build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G "Visual Studio 17 2022" -T ClangCL -DBUILD_BENCHMARKS=ON #TODO: -DWARNINGS_AS_ERRORS=ON
```
rldoris, line 57: Do we want benchmarks on by default?

riley-kinahan:
Most of the changes look good to me. Main thing I'm not sure on is the switch to string views.

I don't understand how the lifetime of string data is managed now. It seems possible to me for a string views within a field container to outlive the actual data leading to a memory access error.

Also had a conversation with @rldoris this week about whether we can improve memory efficiency.  IIRC we have something like a 30x memory usage rate vs raw binary. One big culprit is that storing basic types like `uint64_t`s in a variant with `std::strings` and `std::vectors` leads to a lot of wasted bytes. Each `FieldContainer` is 64 bytes as of right now. Switching to `std::string_view`s gets us down to 56, but it might be worth considering using more primitive types for further savings.

python/test/test_novatel_types.py
```diff
@@ -90,49 +90,33 @@ def test_ascii_char_byte_valid(helper):
     helper.create_base_field("INT_4", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
     helper.create_base_field("INT_5", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
     helper.create_base_field("INT_6", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
-    helper.create_base_field("INT_7", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
-    helper.create_base_field("INT_8", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
-    helper.create_base_field("INT_9", FIELD_TYPE.SIMPLE, "%B", 1, DATA_TYPE.CHAR)
 
-    input = b"-129,-128,-127,-1,0,1,127,128,129"
+    input = b"-128,-127,-1,0,1,127"
```
rldoris, line 94: I think the purpose of this was to document what happens if a larger value somehow gets passed into a field. Should we not be testing/documenting what happens when this happens?

src/decoders/common/src/message_decoder.cpp
```diff
@@ -26,12 +26,10 @@
 
 #include "novatel_edie/decoders/common/message_decoder.hpp"
 
-#include <nlohmann/json.hpp>
+#include <simdjson.h>
```
rldoris, line 29: Do we know the limits of this library? e.g. will it work on a raspberry pi or other small arm processors?

.github/workflows/build_cpp.yml
jonathanmcdermid: we need to build the benchmarks somewhere in the pipeline otherwise people are free to break in PRs, which has already been happening

.github/workflows/build_python.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           # - [Windows, windows-latest, x86]
           # - [Windows, windows-latest, ARM64]
           # - [MacOS, macos-latest, x86_64]
-          - [MacOS, macos-latest, arm64]
+          # - [MacOS, macos-latest, arm64]
```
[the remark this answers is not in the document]
jonathanmcdermid: This is an issue that I am not certain how to solve. This is found in pyproject.toml
`
# Needed for full C++17 support
MACOSX_DEPLOYMENT_TARGET = "13.0"
`
But this just doesnt seem to be the case. Using from_chars on floats is supposed to be enabled in C++17, but it fails for this build configuration. I tried updating the deployment target but it still had the same issue.

examples/novatel/converter_parser/converter_parser.cpp
```diff
@@ -124,9 +118,13 @@ int main(int argc, char* argv[])
 
     while (!ifs.eof())
     {
+        std::array<char, MAX_ASCII_MESSAGE_LENGTH> cData;
         ifs.read(cData.data(), cData.size());
         clParser.Write(reinterpret_cast<unsigned char*>(cData.data()), ifs.gcount());
 
+        MetaDataStruct stMetaData;
+        MessageDataStruct stMessageData;
```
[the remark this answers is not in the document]
jonathanmcdermid: I think either solution is fine, and the performance cost is pretty much nothing for trivial constructing/destructing basic structs like this. The reason I made this change was just to clarify that these members are completely overwritten for each cycle. what do you think?

python/test/test_novatel_types.py
jonathanmcdermid: we could but I certainly wouldnt have this in a test for valid data

src/decoders/common/src/message_decoder.cpp
jonathanmcdermid: https://openbenchmarking.org/result/2202104-NE-RASPBERRY71
https://lemire.me/blog/2019/08/01/a-new-release-of-simdjson-runtime-dispatching-64-bit-arm-support-and-more/
it looks like it works on both

cmake/CompilerOptions.cmake
```diff
@@ -1,7 +1,7 @@
 if(MSVC)
     set_property(GLOBAL PROPERTY USE_FOLDERS ON)
     add_definitions(-DWIN32 -D_WINDOWS)
-    add_compile_options(/W4 /GR /EHsc /utf-8 /wd4244 /wd4293 /wd4996)
+    add_compile_options(/W4 /GR /EHsc /utf-8 /wd4244 /wd4293 /wd4702 /wd4996)
```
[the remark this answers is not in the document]
jonathanmcdermid: I can add comments for that. basically MSVC warnings kinda suck and need to be disabled sometimes.

conanfile.py
valgur: Needs `transitive_headers=True, transitive_libs=True` as well, since the library is used in a public header (include/novatel_edie/decoders/common/message_decoder.hpp).

.github/workflows/build_python.yml
valgur: Python 3.9 fails in tests with
```
          for models in validmodels.models:
  >           assert models.model == "FFNRNNCBN"
  E           AssertionError: assert 'FFNR\x00\x00\x00\x00\x00' == 'FFNRNNCBN'
  
  /project/python/test/test_decode_encode.py:628: AssertionError
  ```
  
  Looks suspicious and worth fixing to me (not necessarily in this PR).

Comment (valgur) [reply .github/workflows/build_python.yml:10]: Yeah, definitely something fishy going on here. The test passes on `main`, but `validationmodels.models[0].model` is all zeroes when I test locally on this branch.

File: conanfile.py (hunk 3 above)
Comment (jonathanmcdermid) [reply conanfile.py:50]: we havent fully dropped it, we still use it for the stuff in json_db_reader. I assume it would be best to replace that stuff with simdjson as well but this PR is already huge.

**Comment by jonathanmcdermid** [comment]
>I don't understand how the lifetime of string data is managed now. It seems possible to me for a string_view within a FieldContainer to outlive the actual data leading to a memory access error.
@riley-kinahan Youve touched on an important limitation here, and its that our fieldcontainer values (not definitions) are only guaranteed to be valid while the message we are working with is still in the buffer and hasnt been overwritten. I think this is a limitation that we can work with, but it should be understood nonetheless.

**Comment by riley-kinahan** [comment]
> > I don't understand how the lifetime of string data is managed now. It seems possible to me for a string_view within a FieldContainer to outlive the actual data leading to a memory access error.
> 
> @riley-kinahan Youve touched on an important limitation here, and its that our fieldcontainer values (not definitions) are only guaranteed to be valid while the message we are working with is still in the buffer and hasnt been overwritten. I think this is a limitation that we can work with, but it should be understood nonetheless.
> 
> Additionally, I feel like this concept was already sort of baked in to the code. We preemptively call clear on the vector of fieldcontainers when we start decoding.

Ah ok, so my understanding is that during parsing messages will become invalid each time the framer buffer is overwritten. As a result it will be unsafe to access a message after the next one has been parsed.

There are a couple Python use cases which this would break, both of which I think are important to support:

1. Saving messages for delayed bulk processing:
```
bestpos_msgs = []
for msg in fp:
     if isinstance(msg, BESTPOS):
        bestpos_msgs.append(msg)
```

2. Multi-threaded processing:
```
while True:
    data = port.listen()
    parser.write(data)
    try:
        msg = parser.read()
        handler.put(msg)
    except ...
```

Do you have an estimate on how much efficiency is gained by avoiding string copying in particular?

File: cmake/CompilerOptions.cmake (hunk 10 above)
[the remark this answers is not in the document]
Comment (rldoris) [reply cmake/CompilerOptions.cmake:4]: Unrelated to your changes, but it would be nice to document it. Whenever I realize I don't have a clue what they are doing it's a pain to look up.

File: src/decoders/common/src/message_decoder.cpp (hunk 7 above)
Comment (rldoris) [reply src/decoders/common/src/message_decoder.cpp:29]: Fantastic. It looks like it runs on mobile processors (iPhone) and PIs so I'm happy with that. I've heard some grumbling about it not supporting embedded processors, but that's a future problem for us to solve (or not solve).

**Comment by rldoris** [comment]
> > Also had a conversation with @rldoris this week about whether we can improve memory efficiency.  IIRC we have something like a 30x memory usage rate vs raw binary. One big culprit is that storing basic types like `uint64_t`s in a variant with `std::strings` and `std::vectors` leads to a lot of wasted bytes. Each `FieldContainer` is 64 bytes as of right now. Switching to `std::string_view`s gets us down to 56, but it might be worth considering using more primitive types for further savings.
> 
> An interesting problem. The best solution I managed to find was a custom variant implementation that automatically allocates all types above some size on the heap: https://cpp-rendering.io/small-object-optimization/ Definitely needs some benchmarking whether the extra indirection hurts the performance noticeably. I would assume it's a net benefit, though, since strings and vectors are not that often encountered in the messages.

@valgur The specific use case that developers are encountering is memory magnification when trying to store the data in Parquet/HD5 formats. It's more efficient to insert all the logs at once so they are writing code like this:

```messages = []
for message in parser.read():
    messages.append(message)
    
// Insert messages into Parquet file
```

and are seeing 32Gb systems run out of memory when processing 1.3Gb files. They've switched to processing the files in chunks, but the amplification is a bit too much so if we can minimize it I would like to try. Ultimately the best solution will be for us to document/provide examples for how to process large datasets because users will always generate large enough datasets that will overwhelm the memory of a PC.

jonathanmcdermid:
@riley-kinahan performance gains from using string_view are significant, as string copying is expensive. I dont have the exact numbers available but it was one of the more significant changes made. I would say that if we need the behaviour your asking for, it would probably be best to do something like add a deep_copy() function for the fieldcontainers when you need them to outlive the lifespan of the message in the buffer, and only call it when needed. What do you think? @rldoris @valgur

rldoris:
I was extremely happy to get rid of the `deepcopy()` functionality on the python side. Whenever I tried to introduce/explain the memory management to users on the python side I very quickly realized that was a big pain point for them. I would definitely cast my vote towards each object having safe/unique memory that doesn't require the user to manage. 

Another way to look at this too is that there aren't a ton of string fields in our most popular logs (VERSION is probably logged once so it doesn't count 😄) so having to copy string value probably isn't going to have a huge performance impact on average.

.github/workflows/build_python.yml
riley-kinahan: @rldoris Do you know if this ever get fixed?

riley-kinahan (approved):

PR #119, opened 2025-03-21
Logging Refactor

riley-kinahan:
Updated logging mechanism to be fully manageable on Python side via the standard library `logging` module.

On the C++ side I refactored the `Logger` static class into a `LoggingManager` singleton class. At runtime the `LoggingManager` singleton instance `pclLoggerManager` will be either a `CPPLoggerManager` or a `PyLoggerManager` depending on whether the Python API has been initialized. Classes register their loggers using the `pclLoggerManager` singleton to allow polymorphic behavior.

```mermaid
classDiagram
    class LoggerManager{
        +RegisterLogger(std::string& name)
        +Shutdown()
    }

    class CPPLoggerManager {
    }

    class PyLoggerManager {
    }

   class Logger {
   }

    LoggerManager<|-- CPPLoggerManager
    LoggerManager<|-- PyLoggerManager
    Logger ..> CPPLoggerManager : wraps
```

C++ users can interact with the `LoggingManager` by using the `GetLoggerManager` function to receive a pointer to the the global manager as a `CPPLoggerManager` instance, or they can do this indirectly using the old `Logger` interface which is remapped to use `GetLoggerManager` under the hood.

```
GetLoggerManager->InitLogger();
or
Logger::InitLogger();
```

Python users can simply access each logger with the `logging.getLogger` function and configure them as they would a regular Python logger. 
```
ne_logger = logging.getLogger('novatel_edie')
decoder_logger = logging.getLogger('novatel_edie.decoder')

ne_logger.setLevel(logging.INFO)
decoder_logger.setLevel(logging.WARNING)
```
Under the hood this is achieved by setting up all C++ spdloggers with a custom sink which sends messages to a corresponding Python logger. 

```mermaid
classDiagram
    class spdlog_logger {
    }

    class base_sink {
    }

    class python_sink {
        -logger: handle
    }
    spdlog_logger o-- base_sink
    base_sink<|-- python_sink 
```

For performance reasons the code also inserts a callback hook into the `setLevel` method of each Python logger which is associated with a spdlogger, allowing the C++ code to cache the Python logger level and avoid costly calls to check the level on each logging attempt. 
Untitled-2024-09-05-1007

Added unit tests included to validate that Python logging interface matches expected behavior.

```
Benchmarking overview for parsing 100k messages:

- Log level too high for anything to be logged
    - With unchanged log attempts: 
        - Old: 3.289s
        - New: 3.119s
    - With additional log per message decoded: 
        - Old: 3.496s
        - New: 3.194s
- Log level allows logging:
    - With unchanged log attempts: 
        - Old: 8.444s
        - New: 6.599s
    - With additional log per message decoded (approximated from 10k messages): 
        - Old: ~700s
        - New: ~400s
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,6 @@ include(cmake/ThirdParty.cmake)
 # Build Project
 project(novatel_edie VERSION ${RELEASE_VERSION} LANGUAGES CXX)
 
-
 option(BUILD_TESTS "Build tests" ON)
 option(BUILD_EXAMPLES "Build examples" ON)
 option(BUILD_BENCHMARKS "Build benchmarks" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,8 @@ include(cmake/ThirdParty.cmake)
 # Build Project
 project(novatel_edie VERSION ${RELEASE_VERSION} LANGUAGES CXX)
 
+add_definitions(-D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR)
+
 option(BUILD_TESTS "Build tests" ON)
 option(BUILD_EXAMPLES "Build examples" ON)
 option(BUILD_BENCHMARKS "Build benchmarks" OFF)
--- a/examples/novatel/command_encoding/command_encoding.cpp
+++ b/examples/novatel/command_encoding/command_encoding.cpp
@@ -43,11 +43,11 @@ int main(int argc, char* argv[])
 {
     // This example uses the default logger config, but you can also pass a config file to InitLogger()
     // Example config file: logger\example_logger_config.toml
-    CPPLoggerManager* pclLoggerManager = GetLoggerManager();
-    std::shared_ptr<spdlog::logger> logger = pclLoggerManager->RegisterLogger("command_encoder");
+    CPPLoggerManager* pclMyLoggerManager = GetLoggerManager();
+    std::shared_ptr<spdlog::logger> logger = pclMyLoggerManager->RegisterLogger("command_encoder");
     logger->set_level(spdlog::level::debug);
-    pclLoggerManager->AddConsoleLogging(logger);
-    pclLoggerManager->AddRotatingFileLogger(logger);
+    pclMyLoggerManager->AddConsoleLogging(logger);
+    pclMyLoggerManager->AddRotatingFileLogger(logger);
 
     if (argc < 3)
     {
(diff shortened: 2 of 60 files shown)


examples/novatel/command_encoding/command_encoding.cpp
```diff
@@ -43,11 +43,12 @@ int main(int argc, char* argv[])
 {
     // This example uses the default logger config, but you can also pass a config file to InitLogger()
     // Example config file: logger\example_logger_config.toml
-    Logger::InitLogger();
-    std::shared_ptr<spdlog::logger> logger = Logger::RegisterLogger("command_encoder");
+    CPPLoggerManager* pclMyLoggerManager = GetLoggerManager();
+    pclMyLoggerManager->InitLogger();
+    std::shared_ptr<spdlog::logger> logger = pclMyLoggerManager->RegisterLogger("command_encoder");
```
rldoris, line 48: Would it be better/easier to define a macro 
`#define CREATE_LOGGER() pclMyLoggerManager->RegisterLogger(std::filesystem::path(__FILE__).stem().string())`

Then we can get similar behaviour `auto logger = CREATE_LOGGER();` but in a hopefully cleaner/easier way or is that obfuscating the functionality too much? I haven't looked, but I expect to see this pattern in the PR quite a bit.

python/examples/common_setup.py
rldoris, line 1: Is this complete? Should we include basic EDIE decoding in here too? I'm unsure what the value of this is.

rldoris: ah you're using it in the examples to remove the inconsistencies in the examples. Nothing to see here. Carry on.

examples/novatel/command_encoding/command_encoding.cpp
riley-kinahan: Good suggestion! 

I also added a `LOGGER_MANAGER` macro to enable usage like `LOGGER_MANAGER->Shutdown()` which is a little cleaner than `GetLoggerManager()->Shutdown()` imo.

rldoris (approved):

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ include(cmake/ThirdParty.cmake)
 # Build Project
 project(novatel_edie VERSION ${RELEASE_VERSION} LANGUAGES CXX)
 
+add_definitions(-D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR)
```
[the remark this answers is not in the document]
riley-kinahan: Comment added in [my latest PR](https://github.com/novatel/novatel_edie/pull/129/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL11)

Thanks for pointing it out, totally slipped my mind!

PR #123, opened 2025-03-24
Separate python and C++ versions and set python version to 2.0.0a

riley-kinahan:
Update CI/CD to be able to handle separate versioning for Python and C++ components. Also set Python version so that it will be incremented to an alpha "2.0.0a" release which can then be published.


valgur:
> @valgur do you think the behavior here is logical and clear?
> 
> Separating out the versions gives us more flexibility to work on the components individually and manage their releases. In particular an alpha version string like `2.0.0a1` is not a valid CMake project version.
> 
> Let me know if you support this change or have any suggestions.

The separate versioning scheme and tags for Python are very reasonable, indeed.

Regarding the alpha pre-release, I suggest using the ([semver recommendation](https://semver.org/#spec-item-9)) of separating the pre-release label with a hyphen, e.g. `2.0.0-a1`. Conan also uses the presence of a hyphen-separated part to interpret a version to be a pre-release, for example.

The new labels for PRs are definitely useful, but perhaps using milestones for that purpose would make more sense (https://github.com/novatel/novatel_edie/milestones) and would also provide a nice overview of a progress towards a release? Did you already consider them?

riley-kinahan:
> > @valgur do you think the behavior here is logical and clear?
> > Separating out the versions gives us more flexibility to work on the components individually and manage their releases. In particular an alpha version string like `2.0.0a1` is not a valid CMake project version.
> > Let me know if you support this change or have any suggestions.
> 
> The separate versioning scheme and tags for Python are very reasonable, indeed.
> 
> Regarding the alpha pre-release, I suggest using the [semver recommendation](https://semver.org/#spec-item-9) of separating the pre-release label with a hyphen, e.g. `2.0.0-a1`. Conan also uses that convention to interpret a version to be a pre-release, for example.
> 
> The new labels for PRs are definitely useful, but perhaps using milestones for that purpose would make more sense (https://github.com/novatel/novatel_edie/milestones) and would also provide a nice overview of a progress towards a release? Did you already consider them?

I agree with sticking to the semver format as close as possible, but in this case it is the [python versioning specification disallows it.](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers:~:text=Semantic%20versions%20containing%20a%20hyphen%20(pre%2Dreleases%20%2D%20clause%2010)%20or%20a%20plus%20sign%20(builds%20%2D%20clause%2011)%20are%20not%20compatible%20with%20this%20specification%20and%20are%20not%20permitted%20in%20the%20public%20version%20field)

I haven't previously used milestones but after reading up on them it seems they are best used to track multiple issues/PRs over a finite time period. I think this makes them less suited to managing versioning on a per-PR basis indefinitely. They do seem like a good tool for organizing work though, [I created one to track the official Python package release!](https://github.com/novatel/novatel_edie/milestone/2)

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "novatel_edie"
 description = "SDK for interfacing with and decoding data from NovAtel OEM7 receivers"
-version = "3.5.0"
+version = "1.0.0"
```
rldoris, line 4: We may want to bump this to v2 so we don't release a version less than what's currently out there
https://pypi.org/project/novatel-edie/ shows that we currently have 1.5.9 released

riley-kinahan: I added a `python major` and `python prerelease` tag to this PR so when it gets merged the python version will undergo a major version increment as a pre-release. E.g. `1.0.0 -> 2.0.0a`

rldoris (approved):

PR #131, opened 2025-04-13
Fix macOS builds, improve Python CI pipeline

valgur:
Adds workarounds for the disappointing `std::from_chars`/`std::to_chars` incompatibilities on floating-point types for macOS and GCC on aarch64:
- use https://github.com/fastfloat/fast_float for `std::from_chars`;
- use `fmt::format` for `std::to_chars`.

Modernize and improve the Python builds:
- switch to the cibuildwheel GitHub action instead of pip-installing and running it;
- get rid of the need for QEMU by using the new `ubuntu-latest-arm` GitHub runner;
- extend `conan_provider.cmake` to support `universal2` architecture (arm64+x86_64) wheels;
- fix Conan incorrectly defaulting to the legacy `libstdc++` ABI instead of `libstdc++11` one.

Python 3.9 support is still missing, though.


riley-kinahan:
Python build changes are really nice!

PR #132, opened 2025-04-14
Improve ascii header tokenization

riley-kinahan:
Problem:
If extraneous delimiters are introduced to a otherwise valid header, the header decoder will still return success, although the actual decoded data will be incorrect. I saw this occur in a dataset which somehow became corrupted with some randomly occurring `NULL` characters. The old `strcspn` function essentially treated these `NULL` characters as valid delimiters.

Solution:
I've updated the `DecodeAsciiHeaderField` function to return failure if it does not encounter the exact delimiter its looking for.

### DIFF
--- a/src/decoders/oem/src/header_decoder.cpp
+++ b/src/decoders/oem/src/header_decoder.cpp
@@ -174,13 +174,10 @@ STATUS HeaderDecoder::Decode(const unsigned char* pucLogBuf_, IntermediateHeader
     {
     case HEADER_FORMAT::ASCII:
         ++pcTempBuf; // Move the input buffer past the sync char '#'
-        if (!DecodeAsciiHeaderFields<pcAsciiRegDelimiter, ASCII_HEADER::MESSAGE_NAME, ASCII_HEADER::PORT,
-                                     ASCII_HEADER::SEQUENCE, ASCII_HEADER::IDLE_TIME, ASCII_HEADER::TIME_STATUS, ASCII_HEADER::WEEK,
-                                     ASCII_HEADER::SECONDS, ASCII_HEADER::RECEIVER_STATUS, ASCII_HEADER::MSG_DEF_CRC>(stInterHeader_, &pcTempBuf))
-        {
-            return STATUS::FAILURE;
-        }
-        if (!DecodeAsciiHeaderField<pcAsciiFinalDelimiter, ASCII_HEADER::RECEIVER_SW_VERSION>(stInterHeader_, &pcTempBuf))
+        if (!DecodeAsciiHeaderFields<pcAsciiRegDelimiter, ASCII_HEADER::MESSAGE_NAME, ASCII_HEADER::PORT, ASCII_HEADER::SEQUENCE,
+                                     ASCII_HEADER::IDLE_TIME, ASCII_HEADER::TIME_STATUS, ASCII_HEADER::WEEK, ASCII_HEADER::SECONDS,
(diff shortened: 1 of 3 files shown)


src/decoders/oem/src/header_decoder.cpp
```diff
@@ -64,15 +65,32 @@ template <ASCII_HEADER eField> bool HeaderDecoder::DecodeAsciiHeaderField(Interm
 
     static_assert(bIsLeadingNumberField || bIsLeadingAlphaField || bIsLeadingHexField);
 
+    size_t ullDelimiterSize = strlen(pcDelimiter);
```
jonathanmcdermid, line 68: should be able to make this constexpr, maybe with sizeof(). If not, could pass a std::array as template arg

src/decoders/oem/src/header_decoder.cpp
```diff
@@ -126,8 +145,12 @@ void HeaderDecoder::DecodeJsonHeader(json clJsonHeader_, IntermediateHeader& stI
 }
 
 // -------------------------------------------------------------------------------------------------------
-STATUS
-HeaderDecoder::Decode(const unsigned char* pucLogBuf_, IntermediateHeader& stInterHeader_, MetaDataStruct& stMetaData_) const
+
+constexpr char pcAsciiRegDelimiter[] = ",";
+constexpr char pcAsciiFinalDelimiter[] = ";";
+constexpr char pcAbbrevAsciiRegDelimiter[] = " ";
+constexpr char pcAbbrevAsciiFinalDelimiter[] = "\r\n";
```
jonathanmcdermid, line 152: if these are only used in decode, declare them within the function

riley-kinahan: Doesn't compile. These are the errors I get:
```
no instance of function template "novatel::edie::oem::HeaderDecoder::DecodeAsciiHeaderFields" matches the argument list
```
```
'novatel::edie::oem::HeaderDecoder::DecodeAsciiHeaderFields': no matching overloaded function found
```

Definitions might need to be accessible to the `DecodeAsciiHeaderField` function?

src/decoders/oem/src/header_decoder.cpp
riley-kinahan: `sizeof` won't work because it will just get the pointer size.

`std::array` is doable, but would have to check equality manually instead of using `strncmp`.

I did confirm by the running a `RELEASE_WITH_DEBUG` build that `ullDelimiterSize` is optimized out by the MSVC compiler.
image

Let me know if you think the change is still worth making.

src/decoders/oem/src/header_decoder.cpp
jonathanmcdermid: make them static within the function

src/decoders/oem/src/header_decoder.cpp
jonathanmcdermid: ignore this suggestion

src/decoders/oem/src/header_decoder.cpp
riley-kinahan: Like this?

image

Doesn't seem to resolve it for me

src/decoders/oem/src/header_decoder.cpp
jonathanmcdermid: that looks preferable

jonathanmcdermid (approved):
looks good!

PR #134, opened 2025-04-17
Improve handling of responses

riley-kinahan:
Problem:
Responses currently are given the same type as regular messages. E.g. a BESTPOS response has a `novatel_edie.messages.BESTPOS` type. This means that the fields don't match the type.

Abbreviated ascii responses are also handled poorly. They aren't decoded into a useable format and can't be encoded.

Solution:
Introduce a `Response` type. Use the `name` field to indicate which message its associated with. 

Make abbreviated ascii responses decode to an empty header (with the response bit set) and set the name as `UNKNOWN`. Encoding a response to abbreviated ascii is also now supported.

### DIFF
--- a/python/bindings/py_message_objects.hpp
+++ b/python/bindings/py_message_objects.hpp
@@ -134,6 +134,7 @@ struct PyMessage : public PyField
 
     PyMessageData encode(ENCODE_FORMAT fmt);
     PyMessageData to_ascii();
+    PyMessageData to_abbrev_ascii();
     PyMessageData to_binary();
     PyMessageData to_flattened_binary();
     PyMessageData to_json();
@@ -147,13 +148,12 @@ nb::object create_message_instance(PyHeader& header, std::vector<FieldContainer>
                                    PyMessageDatabase::ConstPtr database);
 
 
-struct PyResponse : public PyField
+struct PyResponse : public PyMessage
 {
-    PyHeader header;
 
     explicit PyResponse(std::string name_, std::vector<FieldContainer> fields_, PyMessageDatabase::ConstPtr parent_db_,
                         PyHeader header_)
-        : PyField(std::move(name_), false, std::move(fields_), std::move(parent_db_)), header(std::move(header_)) {};
+        : PyMessage(std::move(name_), false, std::move(fields_), std::move(parent_db_), std::move(header_)) {};
 };
 
 } // namespace novatel::edie::oem
--- a/python/bindings/py_message_objects.hpp
+++ b/python/bindings/py_message_objects.hpp
@@ -160,7 +160,7 @@ nb::object create_message_instance(PyHeader& header, std::vector<FieldContainer>
 
 //============================================================================
 //! \class PyResponse
-//! \brief A python representation for a single fully decoded message.
+//! \brief A python representation for a single decoded message response.
 //============================================================================
 struct PyResponse : public PyEncodableField
 {
--- a/python/bindings/py_message_objects.hpp
+++ b/python/bindings/py_message_objects.hpp
@@ -1,11 +1,9 @@
 #pragma once
 
-#include "novatel_edie/decoders/oem/filter.hpp"
-
-
 #include "bindings_core.hpp"
-#include "py_database.hpp"
 #include "message_db_singleton.hpp"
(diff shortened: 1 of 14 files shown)


python/bindings/py_message_objects.hpp
```diff
@@ -121,31 +119,70 @@ struct PyUnknownMessage
...
     PyMessageData to_binary();
     PyMessageData to_flattened_binary();
     PyMessageData to_json();
 };
 
+//============================================================================
+//! \class PyMessage
+//! \brief A python representation for a single fully decoded message.
+//============================================================================
+struct PyMessage : public PyEncodableField
```
rldoris, line 147: Should we be using Field in the base name as that collides with the fields in logs. I'm not sure if this is intentional or not, but would it make more sense to call it PyEncodableMessage or am I misunderstanding the intent of this class?

src/decoders/oem/src/encoder.cpp
```diff
@@ -360,10 +360,24 @@ Encoder::Encode(unsigned char** ppucBuffer_, uint32_t uiBufferSize_, const Inter
...
     }
 
-    STATUS eStatus = EncodeHeader(&pucTempEncodeBuffer, uiBufferSize_, stHeader_, stMessageData_, eHeaderFormat_, eFormat_);
-    if (eStatus != STATUS::SUCCESS) { return eStatus; }
+    if (eFormat_ == ENCODE_FORMAT::ABBREV_ASCII && ((stHeader_.ucMessageType & static_cast<uint8_t>(MESSAGE_TYPE_MASK::RESPONSE)) != 0))
+    {
+        if (!PrintToBuffer(reinterpret_cast<char**>(&pucTempEncodeBuffer), uiBufferSize_, "<")) { return STATUS::BUFFER_FULL; }
+        stMessageData_.uiMessageHeaderLength = 1;
+
+        std::string sResponse = std::get<std::string>(stMessage_[1].fieldValue);
```
rldoris, line 368: Is `[1]` a magic number?

src/decoders/oem/src/parser.cpp
```diff
@@ -135,15 +135,10 @@ Parser::ReadIntermediate(MessageDataStruct& stMessageData_, IntermediateHeader&
...
+            if (stMetaData_.bResponse && stMetaData_.eFormat == HEADER_FORMAT::ABB_ASCII && bMyIgnoreAbbreviatedAsciiResponse)
             {
-                stMessageData_.uiMessageHeaderLength = 0;
-                stMessageData_.pucMessageHeader = nullptr;
-                stMessageData_.uiMessageBodyLength = 0;
-                stMessageData_.pucMessageBody = nullptr;
-                stMessageData_.pucMessage = pucMyFrameBufferPointer;
-                stMessageData_.uiMessageLength = stMetaData_.uiLength;
-                return STATUS::SUCCESS;
+                pclMyLogger->debug("Abbreviated ascii response ignored");
```
rldoris, line 140: You're positive that the stMessageData_ elements are going to be correct without explicitly setting them? e.g. they are getting set earlier in the code when dealing with abbrev. ascii responses.

src/decoders/oem/src/encoder.cpp
riley-kinahan: Yep, the ID is always the 0th field and the string is always the 1st field.

src/decoders/oem/src/parser.cpp
riley-kinahan: Rather than directly setting them here, they'll be set during header and body decoding, following the approach used for all other messages/responses.

python/bindings/py_message_objects.hpp
riley-kinahan: I didn't call it `EncodableMessage` because all "Messages" are encodable. The point of this class is just to act as a place for functionality which is shared between `Message`s and `Response`s.

Whether a `Response` is a `Message` under our spec is ambiguous IMO. They can be modeled very similarly but I decided I liked the freedom of keeping them fully separate within the python interface. The internal CPP hierarchy is less important as it can be refactored without altering any observable behavior.

rldoris (approved):

PR #135, opened 2025-04-23
Add methods for getting values and fields

riley-kinahan:
Adds methods for getting an order list of a message/subfields fields and values. This was a request from a user who wanted a more efficient generic storage format than a dictionary.


python/bindings/py_message_objects.cpp
```diff
@@ -165,6 +165,21 @@ nb::dict& PyField::get_fields() const
     return cached_fields_;
 }
 
+nb::tuple PyField::get_ordered_field_names() const
```
rldoris, line 168: Would ordered be implied and we could just drop it from name & values?

riley-kinahan: I don't think I understand the question. Is this just regarding the name?

python/examples/converter_fileparser.py
```diff
@@ -62,6 +62,9 @@ def main():
     for message in file_parser:
         # Handle messages that can be fully decoded
         if isinstance(message, Message):
+
```
rldoris, line 65: Cleanup extra spaces

python/examples/converter_fileparser.py
```diff
@@ -73,6 +76,9 @@ def main():
                 # Access specific fields
                 lat = message.latitude
                 lon = message.longitude
+            elif isinstance(message, RANGE):
```
rldoris, line 79: Do something if RANGE?

rldoris (approved):

PR #137, opened 2025-05-02
Dev zarnold framer manager v6

zainarnoldhexagon:
Changes to Framer Manager to maintain backwards compatibility


include/novatel_edie/decoders/oem/framer.hpp
```diff
@@ -44,31 +44,59 @@ class Framer : public FramerBase
...
   public:
+    //----------------------------------------------------------------------------
+    //! \brief Reset the state of the Framer.
+    //----------------------------------------------------------------------------
+    void ResetState() override { eMyFrameState = NovAtelFrameState::WAITING_FOR_SYNC; };
+
+    //----------------------------------------------------------------------------
+    //! \brief Initialize the attributes of the Framer.
+    //----------------------------------------------------------------------------
+    void InitAttributes() override
```
rldoris, line 66: Why is this required?

include/novatel_edie/decoders/oem/parser.hpp
```diff
@@ -54,7 +54,7 @@ class Parser
 
     MessageDatabase::Ptr pclMyMessageDb;
     Filter::Ptr pclMyUserFilter;
-    Framer clMyFramer;
+    std::shared_ptr<oem::Framer> pclMyFramer;
```
rldoris, line 57: Is this required for the FramerManager addition? I don't think we need to change the framers to be shared in the parsers and we can keep them unchanged. We may want to switch them over to the FM if we break apart the OEM framer or add other Framers like NMEA, but for now we can just leave existing code alone.

include/novatel_edie/common/circular_buffer.hpp
```diff
@@ -37,7 +37,7 @@
 class CircularBuffer
 {
   private:
-    std::unique_ptr<unsigned char[]> pucMyBuffer; //!< Data buffer (circular buffer)
+    std::shared_ptr<unsigned char[]> pucMyBuffer; //!< Data buffer (circular buffer)
```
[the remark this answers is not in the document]
zainarnoldhexagon: artifact from a previous python-side error. It is back to unique!

include/novatel_edie/decoders/oem/framer.hpp
zainarnoldhexagon: A fix for certain edge cases relating to the issue we were talking about within the OEM framer, where it no longer continues to loop within itself when it gets a false positive. 

There were instances where, once we entered back into the framers, there were artifacts left from the previous run.

I anticipate this will disappear when we separate the OEM Framers into Binary, ASCII, ABBREV etc, since that pattern will have less opportunities to enter these types of edge cases. For now, with the solution we have, it is an unfortunate requirement.

include/novatel_edie/decoders/oem/parser.hpp
zainarnoldhexagon: It needs it for the rxconfighandler. The shared ptr is passed during construction to the rxconfighandler

PR #138, opened 2025-05-08
Update readme and examples

riley-kinahan:
Changes:
- Updated readme and example files.
- Changed some references to a message `body` to refer to its `payload` for more term consistency.
- Added interface instability warnings to RXConfigHandler and RangeDecompressor
- Changed from "default" database to "builtin" database to allow flexibility down the road for making the default database set-able


python/readme.md
```diff
@@ -37,164 +74,352 @@ flowchart TD
...
-    print(ascii_msg)
+for msg in parser:
+    if isinstance(msg, Message):
+        ascii_msg = msg.to_ascii()
+        encode_format_msg = msg.encode(encode_format)
 ```
-Note that any messages which cannot be fully decoded will simply be ignored under this method.
 
-## Message Objects
+Messages can also be transformed into other representations, such as a dictionary via the `to_dict()` method.
```
rldoris, line 251: The `()` is not appearing in the readme.

python/readme.md
```diff
@@ -37,164 +74,352 @@ flowchart TD
...
+### UnknownBytes
+
+`UnknownBytes` are created when the parser determines that a sequence of bytes is undecodable. The type is merely a container for raw byte data.
+
+```tree
+unknown_bytes
+└── data
+```
+
+### MessageData
```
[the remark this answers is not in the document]
rldoris: It might be more clear to say "for the raw bytes of a message/response". How would someone get MessageData back from the parser?

python/readme.md
rldoris, line 1: I think it would be important to have a section that tells the user to isinstance to get message field support in the IDE.

PR #144, opened 2025-05-21
Set python version to 2.0

riley-kinahan:
Mark python version to indicate 2.0 release

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "novatel_edie"
 description = "SDK for interfacing with and decoding data from NovAtel OEM7 receivers"
-version = "2.0.1a4"
+version = "2.0.0a5"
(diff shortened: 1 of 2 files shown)


riley-kinahan:
@valgur Wanted to give you a heads up that we've decided to officially release the python package. Based on the feedback we've gotten internally we're confident that the core interface is where it needs to be. We do have some long term ideas for changes which may be breaking, but we'll be following standard semantic versioning principles going forward so those will be reserved for a 3.0 release.

Thank you for all the work you've contributed to this milestone! The integration of nanobind has allowed for a massive improvement in our interface and made the project much easier to grow and maintain. I don't think we would have ever pursued this approach on our own and your contributions saved us a ton of legwork. The project as a whole is in a much better state thanks to your efforts and we hope to continue working with you going forward!

PR #149, opened 2025-06-16
Fix global variable access pattern

riley-kinahan:
Problem:

Many files accessed the global `pclLoggerManager` directly in order to register loggers. This caused linker errors when building the project with shared libraries.

Solution:

Encapsulate `pclLoggerManager` so that it only every accessed within a single C++ file.

Note:

There still seems to be an issue with some unit tests failing when the project is built with shared libs. It appears to be an issue with spdlog loggers not being gettable from the global registry during unit tests. This is something that may need to be resolved in a future PR.

### DIFF
--- a/scripts/compare_benchmarks.py
+++ b/scripts/compare_benchmarks.py
@@ -25,7 +25,7 @@ with open(summary_path, "a") as summary:
         if current_time is not None:
             diff_pct = ((current_time - main_time) / main_time) * 100
             symbol = "✅"
-            if diff_pct > 50:
+            if diff_pct > 10:
                 symbol = "❌"
                 print(f"::error::Benchmark '{name}' is {diff_pct:.2f}% slower than main")
                 failed = True


src/common/src/logger.cpp
```diff
@@ -1,7 +1,19 @@
...
 std::unique_ptr<LoggerManager> pclLoggerManager = std::make_unique<CPPLoggerManager>();
 
+LoggerManager::~LoggerManager() = default;
+
 CPPLoggerManager* GetLoggerManager() { return dynamic_cast<CPPLoggerManager*>(pclLoggerManager.get()); }
+
+LoggerManager* GetBaseLoggerManager()
+{
+    if (pclLoggerManager == nullptr) { throw std::runtime_error("GetBaseLoggerManager(): LoggerManager is not initialized"); }
+    return pclLoggerManager.get();
```
valgur, line 12: I would recommend initializing `pclLoggerManager` lazily, since initialization of global static is somewhat fragile: https://en.cppreference.com/w/cpp/language/siof.html
The standard guarantees that static initialization is handled in a thread-safe manner: https://timsong-cpp.github.io/cppwp/n3337/stmt.dcl#4
```suggestion
    if (pclLoggerManager == nullptr) { pclLoggerManager = std::make_unique<CPPLoggerManager>(); }
    return pclLoggerManager.get();
```

PR #151, opened 2025-06-18
Generalize message source enum field to sibling_id int field

riley-kinahan:
The [measurement source](https://docs.novatel.com/OEM7/Content/Messages/Binary.htm?tocpath=Commands%20%2526%20Logs%7CMessages%7C_____3#:~:text=Bits%200%2D4%20%3D-,Measurement%20source,-1) header field is not exclusively used for specifying message source in all logs. The available `MEASUREMENT_SOURCE` values are not expansive enough and not necessarily reflective of all values that may be contained in the field.  

This PR updates the `MetaDataStruct` class to use a more generic ucSiblingId `uint8_t` field instead of the old eMeasurementSource `MEASUREMENT_SOURCE::PRIMARY` field. The message name filter is accordingly updated to receive a `uint8_t` value instead of a `MEASUREMENT_SOURCE` value.

The `MEASUREMENT_SOURCE` class enum continues to exist and users may cast the `ucSiblingId` value to it if they want it in that format. In the context of RANGE logs this value can still be treated as a `MEASUREMENT_SOURCE` so when used in RANGECMP specific code it is casted back to a `MEASUREMENT_SOURCE` value.

`MEASUREMENT_SOURCE` also continues to be available in python as an `IntEnum` class. The `source` property of `MessageType` still exists and returns a `MEASUREMENT_SOURCE` but logs a deprecation warning when doing so. The preferred method for accessing this data is now the `sibling_id` property which the user can convert to a `MEASUREMENT_SOURCE` if they wish.

Currently this PR is breaking in C++ due to the changed filter interface and the changes to the `MetaDataStruct` object, and is breaking in Python only on a techincality. Would appreciate suggestions on how to further reduce breakage and how this should be versioned.

Python breaking changes:
- `Filter.message_names` and `Filter.message_ids` return lists of `(string, IntEnum, int)` now instead of  `(string, IntEnum, IntEnum)`. (This is minor and fine to do in a patch release IMO)

C++ breaking changes:
- `MetaDataStruct::eMeasurementSource` has been removed in favour of `MetaDataStruct::ucSiblingId`. (Keeping the old one will increase complexity of keeping the values synced)
- `Filter::IncludeMessageId`, `Filter::RemoveMessageId`, `Filter::IncludeMessageName`, and `Filter::RemoveMessageName` all take a `uint8_t` instead of `MESSAGE_TYPE`. (Could be fixed with overloading)


src/decoders/common/src/message_database.cpp
```diff
@@ -57,14 +57,14 @@ void MessageDatabase::RemoveEnumeration(std::string_view strEnumeration_)
 //-----------------------------------------------------------------------
 uint32_t MessageDatabase::MsgNameToMsgId(std::string sMsgName_) const
 {
-    uint32_t uiSiblingId = 0;
+    uint32_t ucSiblingId = 0;
```
rldoris, line 60: ui is correct because this is an unsigned int/32bit value. I don't think this needs to be changed.

riley-kinahan: Good catch

rldoris (approved):

PR #152, opened 2025-06-19
Change encode function for "e" format string to expect double instead of float

riley-kinahan:
Problem:
The "e" format string is always associated with the `double` data type in novatel message definitions (this isn't explicitly stated anywhere so it should probably work for doubles and floats), however the function for encoding "e" formatted strings was expecting a `float`. This lead to a 'bad variant access' error being raised when encoding certain messages.

Solution:
Updated the encode function to accept either a `float` or `double` type.


src/decoders/oem/src/encoder.cpp
```diff
@@ -143,7 +143,7 @@ void Encoder::InitFieldMaps()
 
     asciiFieldMap[CalculateBlockCrc32("e")] = [](const FieldContainer& fc_, char** ppcOutBuf_, uint32_t& uiBytesLeft_,
                                                  [[maybe_unused]] const MessageDatabase& pclMsgDb_) {
-        return WriteFloatToBuffer(ppcOutBuf_, uiBytesLeft_, std::get<float>(fc_.fieldValue), std::chars_format::scientific, fc_.fieldDef->precision);
+        return WriteFloatToBuffer(ppcOutBuf_, uiBytesLeft_, std::get<double>(fc_.fieldValue), std::chars_format::scientific, fc_.fieldDef->precision);
```
rldoris, line 146: On the decode side we handle both 4 and 8 byte cases
https://github.com/novatel/novatel_edie/blob/67d48bb3d3e0ab125feb520900913f0b1b398a6e/src/decoders/common/src/message_decoder.cpp#L80

Can you find any 4 byte "e" conversion strings or should we remove that case from the decoder?

rldoris (approved):

PR #159, opened 2025-07-08
Fix benchmarking build

riley-kinahan:
It appears that building the benchmarks for the project has suddenly started to fail. The cause appears to be an ambiguity between the single argument and multi argument versions of `DecrompressRangeCmp`. My best guess is that there was an update to the benchmark library which broke us.

To fix this I simply renamed one of the methods.


riley-kinahan:
Benchmarking job will always fail because it can't build on main. I am merging this in to resolve.

PR #165, opened 2025-08-14
Allow filtering of Responses

riley-kinahan:
Add the ability to include or exclude both response and non-response messages at the filtering step. This can help avoid decoding regular messages while waiting for a response or vice-versa.


src/decoders/oem/src/filter.cpp
```diff
@@ -212,6 +212,9 @@ void Filter::ClearFilters()
 
     bMyIncludeNmea = false;
     vMyFilterFunctions.clear();
+
+    bMyIncludeNonResponses = true;
```
rldoris, line 216: Would a nonresponse just be all messages and commands? I'm finding it a bit confusing to have both NonResponses and Responses. If non-responses is just all logs and commands then it would be more clear to say "IncludeMessages"

rldoris (approved):

PR #167, opened 2025-09-03
Replace hard crash with exception

riley-kinahan:
Constructing a `MessageDatabase` using a database file with an invalid RXConfig definition in currently leads to a hard crash in due to the `PyMessageDatabase` constuctor being marked as noexcept. This pull request fixes this by removing the noexcept qualifier and explictly raising a descriptive `FailureException`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,14 @@ edie.*.log
 edie.log
 dist/
 venv/
+.venv/
 *.egg-info
 __pycache__/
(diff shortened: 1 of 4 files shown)


rldoris (approved):

rldoris (approved):
I see the Benchmark failing. We need to review this as we have no way of controlling what machine its running on and how to handle different CPUs performing differently. This is a future us problem.

PR #168, opened 2025-09-04
Add support for 8-byte hex fields

rldoris:
Adding support for a new field type 8-byte hex fields. We currently have support for 1,2, and 4-byte hex fields, but there is a need for 8-byte too.


src/decoders/common/src/message_decoder.cpp
```diff
@@ -69,6 +69,7 @@ void MessageDecoderBase::InitFieldMaps()
     asciiFieldMap[CalculateBlockCrc32("lu")] = SimpleAsciiMapEntry<uint32_t>();
     asciiFieldMap[CalculateBlockCrc32("llu")] = SimpleAsciiMapEntry<uint64_t>();
     asciiFieldMap[CalculateBlockCrc32("lx")] = SimpleAsciiMapEntry<uint32_t, 16>();
```
riley-kinahan, line 71: On the decode side it seems like we have special handling for `"lx"` but  not `"x"` whereas on the encode side we have special handling for `"x"` but not `"lx"`. I think we should be treating both identically in both places.

rldoris: I think the code is this way because x doesn't always mean 1 byte. When explicit size is used `l` or `ll` then we can assume it's that size, but otherwise we need to have a 2nd check. Jon probably added this because there are cases of these bad definitions in the JSON.

riley-kinahan (approved):

PR #169, opened 2025-09-10
Fix cmakepresets causing a infinate loop in MSVC

rldoris:
I would like to switch over to using cmakepresets.json vs cmakesettings.json.
The official MS documentation recommends this and it makes it easier to open other projects when I don't need to turn cmakesettings off/on.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,53 @@
+{
+  "version": 6,
+  "cmakeMinimumRequired": {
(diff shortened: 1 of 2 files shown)


rldoris:
@valgur I don't fully understand the cmake/conan build system so I'm flying a bit blind here. Let me know if there is a better way of achieving this.

riley-kinahan:
This has been resolved by #199

PR #175, opened 2025-09-30
Change fmt version from 11.0.2 to 11.1.3 in conanfile

seanovens:
- Fixes pipeline error caused by string overflow warning in GCC 14. See issue in fmt library [here](https://github.com/fmtlib/fmt/issues/4125)

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -52,7 +52,7 @@ class NovatelEdieConan(ConanFile):
         self.requires("spdlog/[>=1.13 <2]", transitive_headers=True, transitive_libs=True, force=True)
         self.requires("gegles-spdlog_setup/[>=1.1 <2]", transitive_headers=True)
         # fmt/11.1.1 is currently not compatible with spdlog https://github.com/gabime/spdlog/issues/3302
-        self.requires("fmt/11.0.2", force=True)
+        self.requires("fmt/11.1.3", force=True)
 
     def build_requirements(self):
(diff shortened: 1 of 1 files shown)


valgur:
Here's what I ended up using for the [`novatel_edie` recipe](https://github.com/conan-kiln/kiln/blob/main/recipes/novatel_edie/all/conanfile.py) on my public Conan repo:
```python
    def requirements(self):
        self.requires("nlohmann_json/[^3.11]", transitive_headers=True)
        self.requires("spdlog/[^1.10]", transitive_headers=True, transitive_libs=True)
        self.requires("gegles-spdlog_setup/[^1.1]", transitive_headers=True)
```
This allows for flexibility in the versions and omitting `fmt` sidesteps any potential conflicts with the version requirement from `spdlog`. It's still available as a transitive dep from `spdlog`.

`fmt` version 12.0.0 also works with the latest release of novatel_edie, btw.

seanovens:
That appears to cause conflicts with the versions of `fmt` and `spdlog` required by `spdlog_setup`, which makes the build fail in our pipeline. Was there something else you had to change to get it to build with those requirements?

valgur:
Ok, feel free to disregard that comment then. This relied on my upcoming alternative implementation of ConanCenter (https://github.com/conan-kiln/kiln), which takes version ranges and conflict avoidance much more seriously.

PR #228, opened 2026-04-01
Variable array lengths, CRC generators, and FileParserBase

seanovens:
- Adds support for 1, 2, or 4-byte field/variable array lengths (previously, array length was always 4 bytes)
- CRC functions now take generator polynomial through a template argument
- CRC code now supports CRC8, CRC16, and CRC64, with either reflected or non-reflected generators, also handled entirely at compile time
- Added a FileParserBase class parameterized by the parser, filter, metadata, and intermediate header types. After implementing the core components (Framer, Decoder, Encoder, Parser) for a new data format, the FileParserBase class makes it extremely simple to add a FileParser as well.


include/novatel_edie/decoders/common/encoder.hpp
```diff
@@ -305,15 +305,38 @@ template <typename Derived> class EncoderBase
...
+                        switch (lenBytes)
+                        {
+                        case 1:
+                            if (!CopyToBuffer(ppucOutBuf_, uiBytesLeft_, static_cast<uint8_t>(vFcCurrentVectorField.size()))) { return false; }
+                            break;
+                        case 2:
+                            if (!CopyToBuffer(ppucOutBuf_, uiBytesLeft_, static_cast<uint16_t>(vFcCurrentVectorField.size()))) { return false; }
+                            break;
+                        case 0: [[fallthrough]]; // By default, if arrayLengthFieldSize is not specified, encode array length using 4 bytes
+                        case 4:
```
rldoris: Because this is the most common use case should it be the first?

include/novatel_edie/decoders/oem/file_parser.hpp
```diff
@@ -41,29 +35,9 @@ namespace novatel::edie::oem {
 //! \class FileParser
 //! \brief Frame, decode and re-encode OEM logs from an InputFileStream.
 //============================================================================
-class FileParser
+class FileParser : public FileParserBase<Parser, Filter, MetaDataStruct, IntermediateHeader>
```
rldoris: For clarity would it read better as `oem::FileParserBase<>, oem::Filter, ...`? 
Then the user isn't having to reference the file path to know where they are.

include/novatel_edie/decoders/common/file_parser.hpp
```diff
@@ -0,0 +1,326 @@
...
+    ParserT clMyParser;
+    std::shared_ptr<std::istream> pclMyInputStream{nullptr};
+
+    [[nodiscard]] bool ReadStream()
+    {
+        std::array<char, MAX_ASCII_MESSAGE_LENGTH> cData{};
+        pclMyInputStream->read(cData.data(), cData.size());
+        size_t ullBytesRead = pclMyInputStream->gcount();
+        const auto uiBytesRead = static_cast<uint32_t>(ullBytesRead);
+        return ullBytesRead > 0 && clMyParser.Write(reinterpret_cast<unsigned char*>(cData.data()), uiBytesRead) == uiBytesRead;
```
rldoris: Would there be a major impact if we switched the `Parser.Write()` size argument to be `size_t`? Do you think it impact users doing other things (reading from a serial port, socket, etc.)?

rldoris (changes requested):

include/novatel_edie/decoders/common/file_parser.hpp
seanovens: In fact, the OEM parser already takes a `size_t` as an argument, so no need to do the cast here. Fixed 👍

rldoris (approved):

PR #232, opened 2026-04-20
Fix array length field alignment

seanovens:
- Align the array length field for variable length and field arrays in both the decoder and encoder (only relevant for BINARY decoding/encoding)
- Set `arrayLengthFieldSize` in JSON DB reader. Use `arrayLengthFieldSize` in both decoder and encoder

### DIFF
--- a/include/novatel_edie/decoders/common/message_decoder.hpp
+++ b/include/novatel_edie/decoders/common/message_decoder.hpp
@@ -156,17 +156,16 @@ class MessageDecoderBase
     //----------------------------------------------------------------------------
     //! \brief Align the binary buffer pointer to the expected type boundary.
     //
-    //! \param[in] field_ A shared pointer to the field definition describing
-    //! the expected data type and length.
-    //! \param[in] pucStart_ The starting pointer of the binary buffer. Used to
+    //! \param[in] align The alignment requirement in bytes.
+    //! \param[in] start The starting pointer of the binary buffer. Used to
     //! calculate the current offset relative to the beginning of the payload.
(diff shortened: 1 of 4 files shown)


include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -262,22 +261,26 @@ class MessageDecoderBase
...
         {
-            // Use the dataType.length to determine number of bytes used to store the size data.
             // Not all Arrays use 4 bytes to store length data.
-
-            const std::size_t lenBytes = arrayDef.dataType.length;
+            // If arrayLengthFieldSize is specified, use that. Otherwise, if it's a FIELD_ARRAY, use the data type length.
+            // For other array types, default to 4 bytes.
+            const std::size_t lenBytes =
+                (arrayDef.arrayLengthFieldSize == 0 ? (arrayDef.type == FIELD_TYPE::FIELD_ARRAY ? arrayDef.dataType.length : 4)
+                                                    : arrayDef.arrayLengthFieldSize);
```
riley-kinahan: If we update the field array definition initialization logic I think we might be able to use `arrayDef.arrayLengthFieldSize` directly.

https://github.com/novatel/novatel_edie/blob/12b275d905ea183a760cb6af0e31a1fd2a229cf0/src/decoders/common/src/json_db_reader.cpp#L127

riley-kinahan (approved):

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Done 👍 - the `arrayLengthFieldSize` field now defaults to 4 when loading the message database

PR #235, opened 2026-04-27
Python message data structure refactor

riley-kinahan:
# Data Structure Refactor Summary

This PR refactors the Python bindings' internal data structure for representing decoded messages and fields. The goal is to improve performance and enable future mutability features.

### Refactored Field Data Model

Old
New

- Introduced `PyFieldArray` class for array fields with weak reference caching for identity preservation
- Replaced eager dictionary population with lazy field resolution via precomputed `FieldNameMap`
- Non-owning fields hold strong references to parents; data stays alive without copies

### Removed Components
- `DecoderTester` internal testing interface removed. This was directly recreating C++ tests on the python side, its proved to require constant maintainance and I believe we will be better served by tests which use public python interfaces.
- `_novatel_internal` module no longer exported
- Removed `get_field_defs()`, `getitem()`, `contains()`, `len()` from PyField (unused)

### Documentation
- Added [python/docs/message_datastructure.md](../python/docs/message_datastructure.md) with architecture diagram

## **INTERFACE CHANGES**

- Accessing a field-array attribute (e.g., `msg.obs`) now returns a `novatel_edie.FieldArray` instead of a Python `list`

---

### Binary Format (Range Messages)

| Test | Old (msg/s) | New (msg/s) | Speedup |
|------|------------:|------------:|--------:|
| Iterate messages | 30,226 | 27,775 | 0.92x |
| Iterate array elements | 13,257 | 17,058 | **1.29x** |
| Access one field per element | 6,567 | 14,315 | **2.18x** |
| Access all fields (10) in element | 4,135 | 6,300 | **1.52x** |

### ASCII Format (Range Messages)

| Test | Old (msg/s) | New (msg/s) | Speedup |
|------|------------:|------------:|--------:|
| Iterate messages | 8,580 | 9,061 | 1.06x |
| Iterate array elements | 5,993 | 7,133 | **1.19x** |
| Access one field per element | 3,959 | 6,517 | **1.65x** |
| Access all fields (10) in element | 2,768 | 4,000 | **1.44x** |

### Analysis

- **Field access is significantly faster**: Single field access shows **2.18x** improvement for binary, indicating individual field lookups are much more efficient. This is driven primarily by using a lazy approach to python value creation instead of creating a full value dict on first access.
- **Array iteration improved**: 19-29% improvement when iterating array elements. This is primarily thanks to the elimination of copy operations.
- **Message-level overhead unchanged**: Message iteration performance is roughly equivalent, confirming the optimization targets field access, not parsing.
- **Gains across formats**: Both binary and ASCII show improvements, with binary seeing larger gains due to higher baseline throughput.


python/src/py_common/implementations/field_objects.cpp
```diff
@@ -111,86 +115,111 @@ PYCOMMON_EXPORT nb::dict& PyField::to_shallow_dict() const
...
 {
-    if (!contains(field_name)) { throw nb::attribute_error(field_name.c_str()); }
-    return to_shallow_dict()[std::move(field_name)];
+    auto it = fieldNameMap_->find(field_name.c_str());
+    if (it == fieldNameMap_->end()) { throw nb::attribute_error(field_name.c_str()); }
+    return resolve_entry(it->second);
 }
 
-PYCOMMON_EXPORT nb::object PyField::getitem(nb::str field_name) const { return to_shallow_dict()[std::move(field_name)]; }
+PYCOMMON_EXPORT nb::object PyFieldArray::getitem(size_t index) const
```
seanovens: Might be worth adding support for negative indices since that was possible with field arrays previously

riley-kinahan: Good catch! I've added support for it
image

seanovens (approved):

PR #236, opened 2026-04-28
Refactor Python Database Ownership Model

riley-kinahan:
# Consolidate Database with Registration-Based Extras

Replaces the two-class database arrangement on main (`py_common::PyMessageDatabaseCore` + `py_oem::PyMessageDatabase`) with a single Python-facing `py_common::PyMessageDatabase` that owns the underlying `MessageDatabase`, its Python type caches, and any registered message-family extras (e.g. OEM `Encoder`, `RxConfigHandler`).

Family-specific components attach via a registration callback instead of being hard-coded into an OEM-specific wrapper class.

## Key Changes

- **Removed** `py_common::PyMessageDatabaseCore` — its Python type caches, definition mappings, `bindToModule`, and family resolution are now on `PyMessageDatabase`.
- **Removed** `py_oem::PyMessageDatabase` — its `Encoder` + `RxConfigHandler` ownership absorbed by the new `py_common::PyMessageDatabase` via the OEM-specific `DatabaseExtras` allocator.
- **Added** `py_common::PyMessageDatabase` — single class holding `shared_ptr<MessageDatabase>` + Python type caches + extras. Bound to Python as `MessageDatabase`.
- **Added** `MessageFamilyRegistration` — replaces the simple `family → nb::handle` map with `{messageType, allocateExtras}`, so families can attach C++ components without modifying py_common.
- **Added** `py_oem::DatabaseExtras` — OEM allocator that produces the `Encoder` and `RxConfigHandler` previously embedded directly in `py_oem::PyMessageDatabase`.
- **Added** public `MessageDatabase::SetMessageFamily(...)` on the C++ base class, so the new wrapper can mutate metadata without inheriting.
- **Changed** parsers / decoders / file_parser / commander / rxconfig_handler / range_decompressor bindings to accept `PyMessageDatabase::Ptr` and forward `wrapper->core()` to the underlying C++ class.

## Benefits

- **Single point of ownership** for a given database. Python users, decoders, parsers, fields, and messages all reference the same `PyMessageDatabase` instance instead of a wrapper-around-a-core split across two namespaces.
- **Family-agnostic core.** OEM-specific components (`Encoder`, `RxConfigHandler`) are no longer hard-coded into a py_oem class. They attach via a registration callback, so future message families can plug in their own extras with no changes to py_common.
- **Reduces responsibility of Decoder and Parser.** They hold a database pointer rather than reconstructing a family-specific wrapper around it.
- **Allows for more powerful generic code to exist in the database class.** In particular this refactor makes it much more feasible to support pythonic message construction. E.g. `BESTPOS(latitude=0, longitude=0)`.

### DIFF
--- a/python/include/py_common/message_database.hpp
+++ b/python/include/py_common/message_database.hpp
@@ -38,10 +38,11 @@ const MessageFamilyRegistration* GetMessageFamilyRegistration(const std::string&
 class PyMessageDatabaseCore : public MessageDatabase
 {
   public:
+    // All python message databases are to be managed via a shared pointer
     static std::shared_ptr<PyMessageDatabaseCore> Create(MessageDatabase message_db = MessageDatabase());
-
     ~PyMessageDatabaseCore();
 
+    // Only allow construction via static Create method
     PyMessageDatabaseCore(const PyMessageDatabaseCore&) = delete;
     PyMessageDatabaseCore& operator=(const PyMessageDatabaseCore&) = delete;
     PyMessageDatabaseCore(PyMessageDatabaseCore&&) = delete;
@@ -107,12 +108,7 @@ class PyMessageDatabaseCore : public MessageDatabase
     void SetMessageFamily(const std::string& messageFamily);
 
     [[nodiscard]] void* GetMessageFamilyExtras() const { return message_family_extras_; }
-    [[nodiscard]] std::shared_ptr<PyMessageDatabaseCore> GetSharedPtr() const
-    {
-        auto self = self_weak_.lock();
-        if (!self) { throw FailureException("PyMessageDatabaseCore is not owned by a shared pointer."); }
-        return self;
-    }
+    [[nodiscard]] std::shared_ptr<PyMessageDatabaseCore> GetSharedPtr() const { return self_weak_.lock(); }
 
     // MessageDatabase overloads
     void Merge(const std::shared_ptr<PyMessageDatabaseCore> other_);
@@ -171,7 +167,7 @@ class PyMessageDatabaseCore : public MessageDatabase
   private:
     explicit PyMessageDatabaseCore(MessageDatabase&& message_db);
 
-    void InitializeAfterConstruction();
+    void Initialize();
     void ResetMessageFamilyExtras();
 
     //-----------------------------------------------------------------------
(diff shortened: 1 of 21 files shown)


riley-kinahan:
Note that this builds on top of #235

python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -63,12 +64,9 @@ py_common::PyMessageData py_oem::PyEncodableField::PyEncode(ENCODE_FORMAT format
     uint32_t buf_size = MESSAGE_SIZE_MAX * 3;
     if (pclRxConfigHandler->IsRxConfigTypeMsg(this->header.usMessageId))
     {
-        status = this->encoderDb->GetRxConfigHandler()->Encode(&buf_ptr, buf_size, this->header, this->fields, message_data, format);
-    }
-    else
-    {
-        status = this->encoderDb->GetEncoder()->Encode(&buf_ptr, buf_size, this->header, this->fields, message_data, this->header.format, format);
+        status = extras.rxConfigHandler->Encode(&buf_ptr, buf_size, this->header, this->fields, message_data, format);
```
seanovens, line 66:
```suggestion
        status = pclRxConfigHandler->Encode(&buf_ptr, buf_size, this->header, this->fields, message_data, format);
```

python/include/py_common/message_database.hpp
```diff
@@ -194,7 +221,9 @@ class PyMessageDatabaseCore : public MessageDatabase
 
     void ResolveBaseType();
 
-    nb::handle base_message_type;
+    const MessageFamilyRegistration* message_family_registration_ = nullptr;
+    void* message_family_extras_ = nullptr;
```
seanovens: Minor suggestion, feel free to disregard: would it make sense to have a common `MessageFamilyExtrasBase` class that just has a virtual destructor, then make `oem::DatabaseExtras` (and any other format extras) inherit from that common base? That way you could change this to a `std::unique_ptr<MessageFamilyExtrasBase>` and you could get rid of `freeExtras` in `MessageFamilyRegistration`

python/src/py_oem/bindings/py_message_objects.cpp
riley-kinahan: Eliminated variable declaration instead

riley-kinahan:
@seanovens I've updated the PR and its description to reflect the updated approach that eliminates cyclic references.

seanovens (approved):

PR #238, opened 2026-05-11
Fix double field format specifiers to prevent encoding errors

HenleyChase:
## Background
Hello! Thanks for your work on this repo. Here's a bug I found while using it and how I fixed it.

While testing with IONUTC messages, I encountered `Exception: std::get: wrong index for variant` when trying to encode to JSON format.

After investigating, I found that some DOUBLE fields across 10 messages were using the `%e` format specifier, which routes to a handler expecting `float` type. Since these fields store `double` values, the encoder throws the exception when calling `std::get<float>()`.

## My Solution
Changed DOUBLE fields in `database/database.json` to use `%le` instead of using `%e`. 

Tested encoding IONUTC messages before and after the fix:
- **Before**: Exception thrown during JSON encoding
- **After**: Successfully encodes with correct double precision values and sends correct response

Thanks again and please let me know if there is anything you would like me to update about this PR.

### DIFF
--- a/src/decoders/oem/src/encoder.cpp
+++ b/src/decoders/oem/src/encoder.cpp
@@ -237,13 +237,20 @@ void Encoder::InitFieldMaps()
 
     jsonFieldMap[CalculateBlockCrc32("e")] = [](const FieldContainer& fc_, char** ppcOutBuf_, uint32_t& uiBytesLeft_,
                                                 [[maybe_unused]] const MessageDatabase& pclMsgDb_) {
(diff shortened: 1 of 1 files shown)


riley-kinahan:
Hi @HenleyChase, thank you for finding this error and identifying the cause!

I've pushed a commit that updates the json encoder logic instead of the database. This approach is more inline with how the ascii encoding logic currently works and makes the fix backward compatible with older alternative databases.

riley-kinahan (approved):

HenleyChase:
Hey @riley-kinahan thanks so much for taking a look! Yeah good call to update encoder as well.

PR #240, opened 2026-05-25
Python Message Database + Message Definition Structs Refactor

riley-kinahan:
## Python Message Database + Message Definition Structs Refactor

This pull request has the following motivations:
1. **Improve the python interface for managing database contents:**
    - Allow users to construct a message definition and all its sub-components purely in python.
    - The underlying C++ structs contain fields which are used for caching derived values, not configuring a definition, the new interface encapsulates these from the python side.
    - All database methods now use copy semantics for passing definitions to and from python. This avoids the possibility of accidental mutation of a value. There was previously a footgun where appending message definitions from one database to another caused enum definitions to be overwritten. This approach makes issues like this impossible.
2. **Allow the owning database for a type to be resolved:**
    - `Message` subtypes such as `BESTPOS` now hold a pointer to the database which created them. This allows their constructors to lookup the information needed to create a proper object.
    - A database and all python types it creates will now share a lifetime. This means that it is always safe to construct an object. Management is outsourced to python's garbage collector.
    - Once a type has been retrieved from a database, either explicitly or via decoding a message, the database that owns it is guaranteed to be "locked".
3. **Clean up C++ message definition structs:**
    - Centralize initialization logic in constructors instead of relying on json parser.
    - Introduce explicit copy constructors and equality operators.
    - Clearly indicate which fields store cached values.
    - Delete completely dead or redundant code.

## Summary of changes

1. **Value-type rework** — `EnumDefinition`, `BaseField` and subclasses, and `MessageDefinition` got proper copy/move semantics, deep `clone()`, structural `operator==`, and explicit `// cached` annotations on derived fields rebuilt via `RebuildCaches()` / `ComputeStringFlags()`. Default-copy aliasing of nested `string_view`s and `shared_ptr<BaseField>`s is gone. Any communication between python and C++ layer now employs copying.
2. **Database lifetime + locking** — `PyMessageDatabase` is now constructed via a `Create()` factory that publishes the wrapper into nanobind's instance map before init runs. The DB supports `lock()` / `is_locked` and installs `tp_traverse` / `tp_clear` GC slots so the cycle between the DB and its dynamically-built Python types is collectable. `MessageDbSingleton::cleanup` is registered with `atexit` to release the built-in DB at interpreter teardown.
3. **Mutation hooks** — `append_messages`, `append_enumerations`, `remove_message`, `remove_enumeration`, `merge`, `clone`, and `message_family` all throw `UnsupportedException` once locked. Consumers (`Parser`, `FileParser`, `Decoder`, `Commander`, `RangeDecompressor`, `RxConfigHandler`, plus `get_msg_type` / `get_enum_type_*`) implicitly lock the DB on first use. `get_builtin_database()` returns a pre-locked DB.
4. **Definition objects in Python** — `EnumDataType`, `EnumDefinition`, `FieldDefinition`, `EnumFieldDefinition`, `ArrayFieldDefinition`, `FieldArrayFieldDefinition`, `MessageDefinition` are now default-constructible, fully mutable, equality-comparable, and have meaningful `__repr__`s.
5. **Reverse lookup: Python type → definition** — new per-DB `enum_type_lookup_` / `field_type_lookup_` / `message_type_lookup_` maps power `Message.__new__` / `Field.__new__`, so a user can construct an empty message instance of a Python type and have it bind back to the right `MessageDefinition` + CRC.
6. `BaseDataType` deleted (dead code), `SimpleDataType::enums` removed (dead code), `EnumField::length` removed (redundant), `MessageDefinition::messageStyle` removed (dead code).

### C++ API

| Symbol | Status | Migration |
| --- | --- | --- |
| `BaseDataType` | Removed | Use `SimpleDataType` (fields merged in). |
| `SimpleDataType::enums` | Removed | Was unused. |
| `EnumField::length` | Removed | Read `field.dataType.length`. |
| `MessageDefinition::messageStyle` | Removed | No replacement. JSON `messageStyle` keys are now ignored. |
| `BaseField::SetConversion("%")` / `"d"` / etc. | Now throws | Provide a valid format with an alpha specifier. |
| Default copy on `BaseField`/`EnumField`/`ArrayField`/`FieldArrayField`/`MessageDefinition`/`EnumDefinition` | Now deep | Code relying on the previous aliasing breaks (almost certainly was a UAF bug). |
| `FieldArrayField::fieldSize` | Auto-computed in constructor | Don't hand-set after construction. |

External code that builds on `BaseDataType`, reads `EnumField::length`, or touches `messageStyle` will not compile.

### Python API

| Symbol | Status | Migration |
| --- | --- | --- |
| `EnummeratorDefinition` | Aliased to `EnumDefinition` in `novatel_edie/__init__.py` | Imports from `novatel_edie.common_bindings.EnummeratorDefinition` will break — use `EnumDefinition`. |
| `EnumFieldDefinition(name=, enumerators=)` | Removed | New: `(name, type, conversion, data_type, enum_id)`. Add the `EnumDefinition` to the DB separately. |
| `EnumFieldDefinition.length` | Removed | Read `field.data_type`. |
| `EnumFieldDefinition.enum_def` | Read-only | It's a cache resolved by the DB. |
| `FieldDefinition(name, type, conversion, length, data_type)` | Removed | New: `(name, type, conversion, data_type)`; `length` derives from `data_type`. |
| `FieldDefinition.width` / `.precision` | Not bound | Caches recomputed by `SetConversion`. |
| `<enum_type>._name` / `<enum_type>._id` attrs on dynamically-created enum types | Removed | Was unused. |
| `BaseDataType` / `SimpleDataType` Python classes | Removed | `DATA_TYPE` enum is sufficient. |
| `FieldArrayFieldDefinition.field_size` | Removed | Cache; populated by the constructor. |
| `MessageDefinition.fields` setter | Newly available | Was read-only. |
| `MessageDatabase.get_msg_def(...)` | Returns a deep copy | Mutations to the result no longer affect the DB; identity checks against in-DB defs fail. |
| `MessageDatabase.clone()` | Locks the source | Mutate before cloning if you need the source mutable. |
| `MessageDatabase.get_msg_type(...)` / `get_enum_type_by_*(...)` | Implicitly lock the DB | Configure the DB before requesting any Python type. |
| `Parser(db)` / `FileParser(path, db)` / `Decoder(db)` / `Commander(db)` / `RangeDecompressor(db)` / `RxConfigHandler(db)` | Implicitly lock `db` | Same advice. |
| `get_builtin_database()` | Returns locked | Clone first if you need to extend the built-in DB. |
| `field.conversion = "bad"` | `AttributeError` | Was `RuntimeError`. |
| `FieldDefinition(conversion="bad")` etc. | `ValueError` | Constructor errors are now `ValueError`. |
| Mutating a locked DB | `UnsupportedException` | New exception class. |
| `MessageDatabase.lock()` / `.is_locked` | Newly available | New public API. |
| `__eq__` on `EnumDataType` / `BaseField` / `MessageDefinition` | Newly available | Structural equality. |
| `MessageDatabase.append_messages([def, ...])` | Defensively deep-clones each `def` | Source's field defs can no longer be mutated through a shared pointer. |

The current python API was il-defined enough that I suspect most of these changes will have no impact. The database locking is the only exception and is a quick fix with `db.clone()`.

## Test coverage

`python/test/test_message_database.py` — 64 tests (parametrized), all green:

- **Definition objects** — for each of the seven Python definition classes: construction with defaults, with mixed/full kwargs, and via `setattr` after default-construction. Plus `field.conversion = bad-string` → `AttributeError` parametrized over `""`, `"d"`, `"%q!"`, `"%5.2"`.
- **Locking** — six scenarios: explicit `lock()` blocks every mutator, `get_builtin_database()` returns locked, `merge` locks both DBs, `clone` locks the source, `get_msg_type` / `get_enum_type_by_name` lock the DB, and constructing each of `Decoder` / `Parser` / `FileParser` / `Commander` / `RangeDecompressor` / `RxConfigHandler` locks the passed DB.
- **Database actions** — built-in enum lookup, append/remove/merge/clone with bidirectional identity assertions across DBs.

### DIFF
--- a/include/novatel_edie/decoders/common/message_database.hpp
+++ b/include/novatel_edie/decoders/common/message_database.hpp
@@ -326,13 +326,14 @@ struct BaseField
     virtual ~BaseField() = default;
 
     //----------------------------------------------------------------------------
-    //! \brief Structural equality. Returns true when `other` has the same
-    //! runtime type and every visible field compares equal.
+    //! \brief Compares the non-cached fields. Returns true when `other` has the
+    //! same runtime type and every non-cached field compares equal.
     //
-    //! `EnumField::enumDef` is intentionally not part of the comparison —
-    //! it's a cache derived from `enumId`, and we want two `EnumField`
-    //! instances that name the same enum to compare equal regardless of
-    //! which database happens to have resolved their cache pointer.
+    //! Cached fields are intentionally excluded from the comparison. For
+    //! example, `EnumField::enumDef` is a cache derived from `enumId`, and we
+    //! want two `EnumField` instances that name the same enum to compare equal
+    //! regardless of which database happens to have resolved their cache
+    //! pointer.
     //----------------------------------------------------------------------------
     [[nodiscard]] bool operator==(const BaseField& other) const { return typeid(*this) == typeid(other) && equalsImpl(other); }
 
@@ -568,10 +569,11 @@ struct MessageDefinition
     const std::vector<BaseField::Ptr>& GetMsgDefFromCrc(spdlog::logger& pclLogger_, uint32_t uiMsgDefCrc_) const;
 
     //----------------------------------------------------------------------------
-    //! \brief Structural equality. Two `MessageDefinition`s compare equal if
-    //! every scalar field matches and every `BaseField::Ptr` in `fields`
-    //! dereferences to a structurally-equal field. Pointer identity is not
-    //! required, so a deep-copied definition compares equal to its source.
+    //! \brief Compares the non-cached fields. Two `MessageDefinition`s compare
+    //! equal if every scalar field matches and every `BaseField::Ptr` in
+    //! `fields` dereferences to an equal field (which likewise compares only
+    //! non-cached fields). Pointer identity is not required, so a deep-copied
+    //! definition compares equal to its source.
     //----------------------------------------------------------------------------
     [[nodiscard]] bool operator==(const MessageDefinition& other) const
     {
--- a/include/novatel_edie/decoders/common/message_database.hpp
+++ b/include/novatel_edie/decoders/common/message_database.hpp
@@ -226,6 +226,47 @@ struct EnumDefinition
 
     using Ptr = std::shared_ptr<EnumDefinition>;
     using ConstPtr = std::shared_ptr<const EnumDefinition>;
+
+    EnumDefinition() = default;
+    ~EnumDefinition() = default;
+
+    //----------------------------------------------------------------------------
+    //! \brief Deep copy. The lookup maps are rebuilt against the new
+    //! `enumerators` vector — a default copy would alias the source's
+    //! enumerator strings via the `string_view` keys, dangling once the source
+    //! is destroyed.
+    //----------------------------------------------------------------------------
+    EnumDefinition(const EnumDefinition& other) : _id(other._id), name(other.name), enumerators(other.enumerators), unknownValue(other.unknownValue)
+    {
+        for (const auto& enumerator : enumerators)
+        {
+            nameValue[enumerator.name] = enumerator.value;
+            valueName[enumerator.value] = enumerator.name;
+            descriptionValue[enumerator.description] = enumerator.value;
+        }
+    }
+
+    EnumDefinition& operator=(const EnumDefinition& other)
+    {
+        if (this == &other) { return *this; }
+        _id = other._id;
+        name = other.name;
+        enumerators = other.enumerators;
+        unknownValue = other.unknownValue;
+        nameValue.clear();
+        valueName.clear();
+        descriptionValue.clear();
+        for (const auto& enumerator : enumerators)
+        {
+            nameValue[enumerator.name] = enumerator.value;
+            valueName[enumerator.value] = enumerator.name;
+            descriptionValue[enumerator.description] = enumerator.value;
+        }
+        return *this;
+    }
+
+    EnumDefinition(EnumDefinition&&) = default;
+    EnumDefinition& operator=(EnumDefinition&&) = default;
 };
 
 //-----------------------------------------------------------------------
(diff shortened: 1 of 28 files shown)


include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -254,16 +306,17 @@ struct BaseField
...
+    std::optional<int32_t> width;     // cached
+    std::optional<int32_t> precision; // cached
+    bool isString{false};             // cached; has a FIELD_TYPE of STRING or conversion string of either %s or %S
+    bool isCsv{false};                // cached; Ascii encoding of this field is comma-separated,
+                                      // true for arrays which are not strings and do not use the %Z or %P conversion strings
     SimpleDataType dataType;
 
     BaseField() = default;
 
+    // TODO: Remove this? A DATA_TYPE should only have one length
```
seanovens: Agreed, should probably be removed

seanovens (approved):
Code looks good to me - I like the idea of the `_owner_db` attribute for definition lookup.

Might be worth adding a quick blurb to the Python README explaining what database "locking" is, and what actions cause a database to become locked.

include/novatel_edie/decoders/common/message_database.hpp
riley-kinahan: Removed it 👍

include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -380,12 +562,91 @@ struct MessageDefinition
...
+    //----------------------------------------------------------------------------
+    //! \brief Structural equality. Two `MessageDefinition`s compare equal if
+    //! every scalar field matches and every `BaseField::Ptr` in `fields`
+    //! dereferences to a structurally-equal field. Pointer identity is not
+    //! required, so a deep-copied definition compares equal to its source.
+    //----------------------------------------------------------------------------
+    [[nodiscard]] bool operator==(const MessageDefinition& other) const
+    {
+        if (_id != other._id || logID != other.logID || name != other.name || description != other.description ||
+            latestMessageCrc != other.latestMessageCrc || fields.size() != other.fields.size())
```
rldoris, line 581: It's possible for 2 messages to have different CRCs and be identical. The way it's computed on the back end is tied to a hierarchy and not just the field definition so if we want to verify structural equality then we might not need to validate crc

e.g.
crc 1:
   f1: hierarchy1 -> Double Latitude
crc 2:
   f1: hierarchy2 -> Double Latitude

rldoris:
Can you explain "get_builtin_database() returns a pre-locked DB". Does "pre-locked" DB mean a unlocked clone of the database? Can the builtin_database be locked?

rldoris:
Why would clone throw an exception on a locked database? Wouldn't it return a deep copied clone of the database?

rldoris:
`MessageDatabase.clone() 	Locks the source`

Why would cloning a database lock the source? I would expect it to return a deep copied clone of the database.

rldoris:
> `MessageDatabase.clone() Locks the source`
> 
> Why would cloning a database lock the source? I would expect it to return a deep copied clone of the database.

I think I see the reason here: https://github.com/novatel/novatel_edie/pull/240/changes#diff-1dd62eece01eeefbb841deb3be297ce85103d79a2122569ea67551da160a4f0dR142
Would we expect the clone of a database to be shallow? I'm not sure if that's obvious to me on the surface (vs having a `clone_shallow()` call or a name that describes the behavior of the clone)

PR #241, opened 2026-05-28
Intermediate Message Representation Refactor

seanovens:
# EDIE Intermediate Message Representation Refactor

This PR reworks EDIE's internal message data structures, primarily aiming to reduce memory consumption.

## Main changes

- Replaces `std::vector<FieldContainer>` with a new `CompositeField` type (see [docs/composite_field.md](https://github.com/novatel/novatel_edie/blob/dev-seanovens-ir_refactor/docs/composite_field.md) for details)
- Simpler field access on C++ side through `CompositeField::GetFieldValue`
- Centralizes binary field alignment handling with a static registry on `MessageDatabase`

## Bug fixes

- Some previous unit tests for RXCONFIG had embedded logs that were missing optional arguments
- OEM string field padding was previously hardcoded into EncoderBase

## Python changes

The following diagram shows how these changes incorporate with the recent Python message data structure refactor:

python

Note that `varFields` contains one `FieldValueVariant` for each variable-sized field in the message definition. Therefore, `varFields.data()` is not moved/resized once a `CompositeField` is initialized. This means it is safe for `PyField`/`PyFieldArray` to hold pointers to this data, assuming the root `PyField` owns the root `CompositeField`.

A `PyFieldArray` points to the `FieldValueVariant` (either `std::vector<CompositeField>` or `FlatFieldArray`) representing its FIELD_ARRAY. A `PyField` representing an element of a FIELD_ARRAY also holds a pointer to the `FieldValueVariant` of its parent `PyFieldArray`, along with its index in the field array. To support flat field arrays, a `PyField` also holds a pointer to the containing field array's definition, allowing it to access the byte size of each element of the field array.

### Memory use

Using the converter_components example, I decoded 100K binary RANGE messages from an OEM data file, saved them in a std::vector (in which I had reserved 100K slots), and used Massif to measure memory consumption. Using the `std::vector<FieldContainer>` data structure, peak heap memory consumption is just over **10GB**:

```
#-----------
snapshot=60
#-----------
time=69243640663
mem_heap_B=10181268107
mem_heap_extra_B=135376189
mem_stacks_B=0
heap_tree=peak
```

Using `CompositeField`, peak heap memory consumption drops to just under **750MB**, i.e. over 92% less memory:

```
#-----------
snapshot=84
#-----------
time=34653128478
mem_heap_B=747190128
mem_heap_extra_B=3482816
mem_stacks_B=0
heap_tree=peak
```

The refactored implementation consumes 7471.9B per RANGE message. The messages contained 166.3 observations on average, giving an average log body size of 166.3*44B = 7317.2B. Therefore, the refactored implementation has around 2.1% space overhead per message.

#### Decoding

Decoding messages is faster for all formats due in part to a significant decrease in memory allocation. Binary decoding has a larger speedup due to fixed field copy optimization. This is most noticeable with large logs consisting of mostly fixed-size fields like RANGE, which is decoded over **100x faster**.

Previous `std::vector<FieldContainer>`:

```
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
DecodeAsciiLog                  1011 ns         1025 ns       746667 logs_per_second=975.239k/s
DecodeAsciiRangeLog            70947 ns        69754 ns        11200 logs_per_second=14.336k/s
DecodeAbbrevAsciiLog             904 ns          900 ns       746667 logs_per_second=1.11132M/s
DecodeAbbrevAsciiRangeLog      70997 ns        69754 ns         8960 logs_per_second=14.336k/s
DecodeBinaryLog                  352 ns          353 ns      1947826 logs_per_second=2.8332M/s
DecodeBinaryRangeLog           16642 ns        16881 ns        40727 logs_per_second=59.2393k/s
```

New `CompositeField`:

```
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
DecodeAsciiLog                   845 ns          854 ns       896000 logs_per_second=1.17029M/s
DecodeAsciiRangeLog            61454 ns        61384 ns        11200 logs_per_second=16.2909k/s
DecodeAbbrevAsciiLog             794 ns          785 ns       896000 logs_per_second=1.27431M/s
DecodeAbbrevAsciiRangeLog      62593 ns        62779 ns        11200 logs_per_second=15.9289k/s
DecodeBinaryLog                  120 ns          120 ns      5600000 logs_per_second=8.33488M/s
DecodeBinaryRangeLog             163 ns          160 ns      4977778 logs_per_second=6.24662M/s
```

#### Encoding

Encoding binary messages is around twice as fast due to fixed field copy optimization. Unfortunately, ASCII encoding is around 15-20% slower. After doing some profiling, my best theory for this is the `FieldValueVariant` allocation/destruction that is performed to do the `asciiFieldMap`/`jsonFieldMap` lookups.

Accessing fixed-size fields that are not aligned according to C++ rules also has some overhead, as bytes must be copied from the fixed region into a properly-aligned value before being returned. (I have not yet attempted to quantify how common such unaligned fields are.)

Previous `std::vector<FieldContainer>`:

```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
EncodeFlattenedBinaryLog       83.1 ns         83.7 ns      7466667 logs_per_second=11.9467M/s
EncodeAsciiLog                  492 ns          500 ns      1000000 logs_per_second=2M/s
EncodeAbbrevAsciiLog            424 ns          414 ns      1659259 logs_per_second=2.41347M/s
EncodeBinaryLog                77.6 ns         76.7 ns      8960000 logs_per_second=13.0327M/s
EncodeJsonLog                   424 ns          424 ns      1659259 logs_per_second=2.35984M/s
```

New `CompositeField`:

```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
EncodeFlattenedBinaryLog       40.2 ns         39.2 ns     17920000 logs_per_second=25.4862M/s
EncodeAsciiLog                  585 ns          586 ns      1120000 logs_per_second=1.70667M/s
EncodeAbbrevAsciiLog            511 ns          500 ns      1000000 logs_per_second=2M/s
EncodeBinaryLog                41.0 ns         40.8 ns     17230769 logs_per_second=24.506M/s
EncodeJsonLog                   516 ns          516 ns      1120000 logs_per_second=1.9373M/s
```

### Python

Memory and speed benefits carry through to the Python API. In the following simple experiment, I took a 4.6GB file of OEM binary data, decoded all of the messages, and saved all of the RANGE messages in a list. Afterwards, to simulate some offline data analysis, I iterated over the saved RANGE messages and added up the "sv_prn" field of every observation.

|                                   | Parsing time (s) | Data analysis time (s)| Peak memory use (GB) |
|-----------------------------------|------------------|-----------------------|----------------------|
| Current `FieldContainer` message  | 47.14            | 121.75                | 50.1                 |
| New `CompositeField`                 | 12.8             | 55.8                  | 4.4                  |

Here are a few additional stats about the test data file:

- Total messages: 3149526
- RANGE messages: 586912 (~19%)
- Other messages: GALINAVRAWWORD (~30%), RAWSBASFRAME2 (~10%), BDSBCNAV2RAWMESSAGE (~10%), other "raw" messages (~10%), ephemeris messages (~10%), LBANDTRACKSTAT (~4%), all others (~7%)

Note that field access efficiency itself is not responsible for the improvement in the data analysis time. Rather, this speedup is probably simply due to greatly reduced memory pressure (and much better cache efficiency).

## Future opportunities

- Redesign `asciiFieldMap`/`jsonFieldMap` to avoid the `FieldValueVariant` copy and generally reduce encoding overhead
- Implement flat field arrays for ASCII/JSON formats for additional space savings (note that any field array can be either `std::vector<CompositeField>` or `FlatFieldArray` and the `CompositeField` will handle it either way)


include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+
+        bool operator==(const const_iterator& other) const { return fields == other.fields && index == other.index; }
+        bool operator!=(const const_iterator& other) const { return !(*this == other); }
+    };
+
+    const_iterator begin() const
+    {
+        if (definition == nullptr) { throw std::runtime_error("begin(): message definition not set"); }
+        return const_iterator(&definition->GetMsgDefFromCrc(defCrc).messageOrderedFields);
+    }
```
riley-kinahan, line 1354: Iteration throwing for nested `MessageBody` objects isn't great. It seems like top-level messages are the primary model while field array objects are treated as a degenerate case. The existing python code takes the opposite approach (which is also a bit messy).

Did you consider making a separate class for the nested values and sharing logic via aggregation or inheritance? The problem I ran into when trying this on the python side is that I needed `Field` objects to be able to work with both `MessageDefintion` and `BaseField` objects. I think the way you've refactored things with both `MessageDefintion` and `FieldArrayField` owning a `FieldInfo` value might allow for a much more elegant solution, where `FieldInfo` can be the core driver of how data is interpreted for both top-level and nested fields.

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+    template <typename T> static FieldValueVariant ReadValueAtPointer(const std::byte* data_, const size_t count_ = 1)
     {
-#ifndef NDEBUG
-        if (!Validate()) { ThrowValidationFailure(); }
-#endif
+        if (data_ == nullptr) { throw std::runtime_error("ReadValueAtPointer(): null data pointer"); }
+
+        if (count_ <= 1)
+        {
+            if (reinterpret_cast<uintptr_t>(data_) % alignof(T) == 0U) { return *reinterpret_cast<const T*>(data_); }
```
riley-kinahan: I think we are better off just using memcpy for this as well

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+            T value{};
+            std::memcpy(&value, data_, sizeof(T));
+            return value;
+        }
+
+        if ((reinterpret_cast<uintptr_t>(data_) % alignof(T)) == 0U) { return reinterpret_cast<const T*>(data_); }
+
+        std::vector<T> values(count_);
+        std::memcpy(values.data(), data_, sizeof(T) * count_);
+        return values;
```
riley-kinahan: Getting a pointer when aligned and a vector otherwise isn't a great interface and there's the reinterpret cast UB issue. I suggest just memcpying the data out. 

The other option would be to return a custom iterator view into the data:

```
template <typename T>
struct TypedBuffer {
    std::byte* data;

    T operator[](std::size_t index) {
        T value{};
        std::memcpy(&value, data + sizeof(T) * index, sizeof(T));
        return value;
    }
};
```

This does still require us to memcpy when an individual value is accessed. There's a C++23 function which fixes the reinterpret_cast UB issue but doesn't work with unaligned data: https://en.cppreference.com/cpp/memory/start_lifetime_as

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+    //! \return true if the field was found and val_ was set, false otherwise.
+    // ---------------------------------------------------------------------------
+    bool GetFieldValue(FieldValueVariant& val_, const std::string& fieldName_) const
+    {
+        if (definition == nullptr) { throw std::runtime_error("GetFieldValue(): message definition not set"); }
+        const auto it = std::find_if(begin(), end(), [&fieldName_](const BaseField& fieldDef) { return fieldDef.name == fieldName_; });
+        if (it == end()) { return false; }
+        val_ = GetFieldValue(*it);
+        return true;
+    }
```
riley-kinahan: Calling this method repeatedly is pretty wasteful. Are we making it too easy to be inefficient here?

I think something like `field.GetFieldValue(field.GetFieldDefByName(name))` would make it clearer what the performance aware approach should be while still making this simple to do when needed.

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+    const MessageDefinition::ConstPtr& GetDefinition() const { return definition; }
+    uint32_t GetDefinitionCrc() const { return defCrc; }
+
+    // ---------------------------------------------------------------------------
+    //! \brief Get a field value by field definition.
+    //!
+    //! \param[in] field_ The definition of the field to retrieve.
+    //! \return A FieldValueVariant containing the field value.
+    // ---------------------------------------------------------------------------
+    const FieldValueVariant GetFieldValue(const BaseField& field_) const
```
riley-kinahan: I can see the value in returning a `FieldValueVariant` as it makes writing generic user code easier, but I think it might be nice to be able to grab out the type you want directly as well:
```
template <typename T>
T GetFieldValue(const BaseField& field_) const;

auto lat = msg.GetFieldValue<float>(latDef);
```

Maybe `FieldValueVariant` could be used as the return type for a value iterator.

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+    // ---------------------------------------------------------------------------
+    //! \brief Get a field value from flat field array storage.
+    //!
+    //! \param[in] fieldDef_ The definition of the field to retrieve.
+    //! \param[in] flatFieldArray_ The flat field array bytes.
+    //! \param[in] index_ The element index within the field array.
+    //! \param[in] fixedFieldBytes_ The byte size of one field array element.
+    //! \return The value of the specified field as a FieldValueVariant.
+    // ---------------------------------------------------------------------------
+    static FieldValueVariant GetValueFromFlatFieldArray(const BaseField& fieldDef_, const std::vector<std::byte>& flatFieldArray_, size_t index_,
```
riley-kinahan: This seems like it could be a private method

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,65 +47,637 @@
...
+        using pointer = BaseField::ConstPtr;
+        using difference_type = std::ptrdiff_t;
+        using iterator_category = std::forward_iterator_tag;
+
+        const std::vector<BaseField::ConstPtr>* fields;
+        size_t index;
+
+        const_iterator(const std::vector<BaseField::ConstPtr>* fields_, size_t index_ = 0) : fields(fields_), index(index_) {}
+
+        reference operator*() const { return *(*fields)[index]; }
```
riley-kinahan: The return value being a `BaseField` instead of a value is not what I would expect as a user. Could be useful but should maybe be under something other than the standard method names.

include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -261,6 +261,8 @@ struct BaseField
     bool isCsv{false};    // Ascii encoding of this field is comma-separated,
                           // true for arrays which are not strings and do not use the %Z or %P conversion strings
     SimpleDataType dataType;
+    size_t index{0}; // if fixed, this is the byte offset of the field in MessageBody.fixedFields
+                     // if variable, this is the index of the field in MessageBody.variableFields
```
riley-kinahan: The change to calculating alignment once at database load time is really nice

docs/messagebody.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Variable region: `std::vector<FieldValueVariant>`
+
+![Variable-length region layout](variable.png)
+
+- Stores vectors of primitive types, strings, and field arrays
+
+#### Field arrays
+
+- Stored as `std::vector<std::byte>` if every field has fixed size
```
riley-kinahan: I think there might be room to refine this a bit. Maybe a wrapper class that implements the getting and setting of fields directly so we don't need a bunch of methods on the parent to do it. 

```
struct SimpleFieldArray {
    std::vector<std::bytes> data;
    MessageDefinition::ConstPtr def;

    ...
}
```

Would cost an extra 8 bytes per `FieldValueVariant` as std::string is 32 bytes on most systems and this would be  40.

riley-kinahan (changes requested):
Really great work on this!

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Good suggestion - I've reworked this so that `MessageBody` (and `FlatFieldArray`) owns a pointer to a `FieldInfo` now rather than a `MessageDefinition`. Field arrays represented as nested `MessageBody` objects have pointers to their sub-`FieldInfo`, so iteration is supported as it is for the top-level message.

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Agreed - got rid of all the unsafe `reinterpret_cast` stuff

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Added a `TypedBuffer` view as you suggested 👍

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Removed that version of `GetFieldValue`, added a `GetFieldDefByName` function to `FieldInfo`, and updated the README with suggested approaches to getting field values.

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Added type-templated overloads for `GetFieldValue` and renamed the variant-returning versions to `GetFieldValueVariant`. Also changed the iterator type of `MessageBody` to `FieldValueVariant` as you suggested.

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Changed to `FieldValueVariant`

docs/messagebody.md
seanovens: Added a `FlatFieldArray` class to make it easier to work with field arrays. Both `MessageBody` and `FlatFieldArray` now use `FieldInfo` rather than `MessageDefinition`.

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Removed this method as I have a separate `FlatFieldArray` class now

python/src/py_common/implementations/message_database.cpp
```diff
@@ -111,7 +111,20 @@ PYCOMMON_EXPORT void py_common::PyMessageDatabase::AppendMessages(const std::vec
...
+        // Rebuild FieldInfo to ensure fixed-length field indices match our message family
+        std::vector<BaseField::Ptr> fieldCopies;
+        for (const auto& [crc, fields] : msgDef->fieldInfo)
+        {
+            std::vector<BaseField::Ptr> fieldCopyVec;
+            for (const auto& f : fields->messageOrderedFields) { fieldCopyVec.push_back(f->clone()); }
+            copy->fieldInfo[crc] = BuildFieldInfo(std::move(fieldCopyVec), core_->GetDbMetadata() ? core_->GetDbMetadata()->messageFamily : "");
+        }
+        owned.push_back(copy);
+    }
```
seanovens: This is an additional change that I wanted to highlight because I'm not 100% sure it's the right behaviour. This rebuilds fieldInfo to ensure all of the added definitions have the right field alignments. So e.g. definitions added to an OEM database would have their field indices recomputed using the OEM alignment function. Adding the same defs to different databases would create new `BaseField` copies with potentially-different `index` values. @riley-kinahan does this break any of the new database semantics? (also, can you think of anywhere I might have missed updating field indices on the Python side of things?)

README.md
```diff
@@ -251,48 +280,69 @@ Run the resulting executable with the following command: `rxconfig_handler.exe <
...
 
 ```cpp
-std::vector<FieldContainer> stMessage;
-double latitude;
+MessageBody stMessage;
 eDecoderStatus = clMessageDecoder.Decode(pucFrameBuffer, stMessage, stMetaData);
 
+// Get a reference to a field definition
+const auto& latDef = stMessage.GetFieldInfo()->GetFieldDefByName("latitude");
+if (!latDef) { throw std::runtime_error("Could not find latitude definition!"); }
```
riley-kinahan: This relies on stMessage being a `BESTPOS` message. Retrieving the field info directly from the database would be preferrable. If there's not a good interface for doing it that way then this is fine to leave as is, but maybe add a todo comment about adding one.

src/decoders/oem/src/message_decoder.cpp
```diff
@@ -190,17 +214,62 @@ void MessageDecoder::InitOemFieldMaps()
...
+        std::string_view msgName;
+        if (clJsonField_.get(msgName) != simdjson::SUCCESS) { throw std::runtime_error("Invalid R JSON value"); }
+        MessageDefinition::ConstPtr pclMessageDef = pclMsgDb_.GetMsgDef(msgName);
+        const auto value = pclMessageDef != nullptr ? CreateMsgId(pclMessageDef->logID, 0, 1, 0) : 0;
+        if (fixed_) { clMessageBody_.SetArrayElement<true>(pstMessageDataType_, elementIndex_, value); }
+        else { clMessageBody_.SetArrayElement<false>(pstMessageDataType_, elementIndex_, value); }
     };
 }
+
+novatel::edie::MessageDefinition::ConstPtr MessageDecoder::GetMessageDefinition(MetaDataBase& stMetaData_) const
```
riley-kinahan, line 246: Nice cleanup moving this response logic out of common

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,67 +47,1245 @@
...
+    //! \param[in] fieldIndex_ The index of the field.
+    //! \param[in] startIndex_ The index in the field's byte region.
+    //! \param[in] value_ The value to set.
+    //! \param[in] n Number of values to copy (defaults to 1).
+    // ---------------------------------------------------------------------------
+    template <typename T, typename = std::enable_if_t<!std::is_pointer_v<T>>>
+    void SetFieldValue(const size_t fieldIndex_, const size_t startIndex_, const T& value_, size_t n = 1)
+    {
+        fields.SetFieldValue((fieldIndex_ * fieldInfo->fixedFieldBytes) + startIndex_, &value_, n);
+    }
```
riley-kinahan: Passing `n != 1` here is always an error

include/novatel_edie/decoders/common/message_decoder.hpp
```diff
@@ -42,67 +47,1245 @@
...
+template <template <typename...> class Template, typename... Args> struct is_specialization_of<Template<Args...>, Template> : std::true_type
+{
+};
+
+template <typename T, template <typename...> class Template> inline constexpr bool is_specialization_of_v = is_specialization_of<T, Template>::value;
 
-#define NOVATEL_TYPES bool, int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t, float, double, std::string
-#define CONTAINER_TYPES NOVATEL_TYPES, std::vector<FieldContainer>
-using FieldValueVariant = std::variant<CONTAINER_TYPES>;
+class MessageBody;
```
riley-kinahan: What do you think of renaming this to `ComplexField`? `MessageBody` implies to me that it can't be nested.

python/src/py_common/bindings/message_database.cpp
```diff
@@ -232,33 +233,62 @@ void py_common::init_common_message_database(nb::module_& m)
...
+            [](FieldArrayField& self) -> const std::vector<BaseField::ConstPtr>& {
+                if (!self.fieldInfo) { self.fieldInfo = std::make_shared<FieldInfo>(); }
+                return self.fieldInfo->messageOrderedFields;
+            },
+            [](FieldArrayField& self, std::vector<BaseField::Ptr> fields, std::string messageFamily) {
+                FieldArrayField rebuilt(self.name, self.type, self.conversion, self.dataType.name, self.arrayLength,
+                                        BuildFieldInfo(std::move(fields), messageFamily));
+                self = std::move(rebuilt);
+            },
+            "fields"_a, "messageFamily"_a = std::string{},
```
riley-kinahan: It doesn't make any sense to have multiple args on a property setter. I don't think we should be reliant on "messageFamily_" being passed in. 

This is the main thing that needs to get resolved before this can be merged, I'll leave my thoughts about how we should be managing the database state in response to your comment about recalculating field alignments.

python/src/py_common/implementations/message_database.cpp
riley-kinahan: This is definitely the right approach but unfortunately I think we need to rework some additional logic to avoid violating the existing python semantics.

The addition of `BaseField::index` means that a `MessageDefintion`'s data is now dependent on which "messageFamily" it is associated with. This means that all `MessageDefinition` objects of a database must be updated when `MessageDatabase.message_family` is updated.

There are two ways this can be done incorrectly:
1. No update is done (current situation): Decoding and encoding silently stop working.
2. Updates are done in place: `MessageDefinition`s from a forked database will be updated in-place and the old database will enter an invalid state.

The way to do it under the current system is for all `MessageDefinition`s to be wiped and replaced with new ones.

FWIW I don't think a `MessageDefinition` storing data dependant on its message family is unreasonable. In principle it should have always been the case because "%P" and "%Z" are novatel specific conversion strings that determine whether `isString` is true, but thus far we've hardcoded those into the common spec out of convenience.

I feel like there's room to improve the design to make things simpler here. The python message database is relying on implementation details of the `MessageDatabase` that are not very well codified. Let me know if you have any thoughts about how we can get to a better solution. I think its ok to leave that to a future PR as long as we can preserve the semantics in this one.

riley-kinahan (commented):
Message database semantics is the only major thing that still needs to get figured out here. Otherwise the changes look great!

docs/messagebody.md
```diff
@@ -0,0 +1,47 @@
...
+### Reading fields
+
+- `MessageBody::GetFieldValue` and `FlatFieldArray::GetFieldValue` can be used to get the value of most fields by their definition
+- `MessageBody::GetFieldValueVariant` wraps a field's value in a variant for programming convenience
+
+### Writing fields
+
+- `MessageBody::SetFieldValue` can be used to set the value of most fields by their definition
+- `MessageBody::SetArrayElement` can be used to set the values of individual elements of FIXED_LENGTH_ARRAYs/VARIABLE_LENGTH_ARRAYs
+- Use `MessageBody::GetFixedFields`/`MessageBody::GetVarFields` to obtain references to the fixed/variable regions, respectively
```
rldoris: Should this be here or is it implying that the references can be written to if the user wants to change the values?

include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -477,23 +479,49 @@ struct ArrayField : BaseField
...
+    size_t varFieldCount{0};
+    std::vector<BaseField::ConstPtr> messageOrderedFields; // vector of field definitions in the order they are encoded in the message
+
+    // ---------------------------------------------------------------------------
+    //! \brief Get a field definition by name.
+    //!
+    //! \param[in] fieldName_ The name of the field definition to retrieve.
+    //! \return A constant pointer to the field definition if found, nullptr otherwise.
+    // ---------------------------------------------------------------------------
+    BaseField::ConstPtr GetFieldDefByName(const std::string& fieldName_) const
```
rldoris, line 494: Should we also include a get GetFieldByIndex? In some tools I think we spend quite a bit of time iterating over the fields looking for the correct name. Having a Idx interface might be good for maximizing performance so it's only doing a bounds check vs iterating + strcmp.

include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -636,12 +681,16 @@ struct MessageDefinition
...
-            auto& copyVec = fields[crc];
-            copyVec.reserve(fieldVec.size());
-            for (const auto& f : fieldVec) { copyVec.push_back(f ? f->clone() : nullptr); }
+            auto copiedFieldInfo = std::make_shared<FieldInfo>();
+            copiedFieldInfo->fixedFieldBytes = otherFieldInfoPtr->fixedFieldBytes;
+            copiedFieldInfo->varFieldCount = otherFieldInfoPtr->varFieldCount;
+            auto& copyVec = copiedFieldInfo->messageOrderedFields;
+            copyVec.reserve(otherFieldInfoPtr->messageOrderedFields.size());
+            for (const auto& f : otherFieldInfoPtr->messageOrderedFields) { copyVec.push_back(f ? f->clone() : nullptr); }
+            fieldInfo[crc] = std::move(copiedFieldInfo);
```
rldoris: This code block is duplicated 3 times in clone() and the copy constructor. Would it be worth turning it into a function to keep the code DRY?

python/src/py_common/implementations/message_database.cpp
seanovens: The first approach that comes to mind is to just copy/rebuild all the message definitions when the message family changes. Slow, but I don't think DB message families will be changed all too often in a typical application (though correct me if I am wrong - perhaps there are some use cases I am not considering)

This is probably something that should be brought down to the C++ layer as well. I propose that copying message definitions is moved down to the C++ `MessageDatabase::AppendMessages` function so that field index computation can be handled entirely in the core database. The C++ `MessageDatabase::SetMessageFamily` function can clear and rebuild the message definitions using `AppendMessages`. I _think_ this would also let `PyMessageDatabase::Merge` use C++'s `MessageDatabase::Merge`, as we wouldn't have to worry about the merged-in DB being corrupted via `MapMessageEnumFields`. Speaking of `Merge` - do you think it makes sense to restrict DB merges to be same-message-family-only, as message definitions can just be shared in that case?

It does feel a bit odd that the indices of a `MessageDefinition` are invalid until the definition is added to a database, but I'm not sure there is a way around that (short of forcing `MessageDefinition` construction to go through the database somehow - but that would be an interface change)

riley-kinahan: Agree with everything here. 

On the weirdness of `MessageDefinition` being filled in after insertion into the DB, I think the core conflict is that we are overloading the `MessageDefinition` struct as both a config source and a runtime data source. These two use cases are similar but want slightly different things.

A `MessageDefinition` and `MessageDefinitionInternal` split might make things a little cleaner but would also add quite a bit of redundancy.

README.md
seanovens: Good point - I've updated that example to get the message definition from the database instead

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Removed the parameter 👍

docs/messagebody.md
seanovens: Right now there there are `Get*Fields` methods for getting either `const` or non-`const` references to the memory regions for convenience. I currently use both for the variable-length fields, though I could probably get rid of the non-`const` version and replace with equivalent calls to `SetFieldValue`.

include/novatel_edie/decoders/common/message_decoder.hpp
rldoris: I don't mind the proposal. I am a bit worried that we now have 3 names for this concept: FieldArray (current), ClassArray (older style), and possibly ComplexField (proposed). Should we consolidate on a single one or do we think we need to poke at it for a bit longer to determine the best name?

include/novatel_edie/decoders/common/message_database.hpp
seanovens: You could achieve this by doing `stMessage.GetFieldValue<T>(stMessage.GetFieldInfo()->messageOrderedFields[i])`, though I can also add a method if that is too verbose.

python/src/py_common/bindings/message_database.cpp
seanovens: Removed `messageFamily` parameters from all properties

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: I think the notion of "field array" remains unchanged, and it is still called a "field array" (there are just two low-level representations of field array now - flat and nested). A "ComplexField" is an individual element of a (nested) field array.

What about "CompositeField"? (To me "ComplexField" sounds like it'd be the opposite of a "Simple" field, but they're sort of independent concepts here.) Also, I suppose "NestedFieldArray" should also be renamed to "Complex/CompositeFieldArray" for consistency?

include/novatel_edie/decoders/common/message_database.hpp
seanovens: Added a `clone` method to `FieldInfo` to reduce repetition here

python/src/py_common/implementations/message_database.cpp
seanovens: Ok I've moved all the definition copying stuff to the C++ layer in `MessageDatabase::AppendMessages` and `MessageDatabase::SetMessageFamily` now uses `AppendMessages` to recompute the definitions. Also added a few unit tests for the C++ `MessageDatabase` in `src/decoders/common/test/message_database_unit_test.cpp`

include/novatel_edie/decoders/common/message_decoder.hpp
seanovens: Renamed `MessageBody` to `CompositeField` and `NestedFieldArray` to `CompositeFieldArray`

src/decoders/oem/test/oem_test.cpp
```diff
@@ -4255,16 +4255,14 @@ class NovatelTypesTest : public ::testing::Test
         STATUS TestDecodeAscii(const std::vector<BaseField::Ptr>& MsgDefFields_, const char** ppcLogBuf_, CompositeField& vIntermediateFormat_)
         {
             const FieldInfo fieldInfo = BuildFieldInfo(MsgDefFields_);
-            vIntermediateFormat_.GetFixedFields() = std::vector<std::byte>(fieldInfo.fixedFieldBytes);
-            vIntermediateFormat_.GetVarFields().resize(fieldInfo.varFieldCount);
+            vIntermediateFormat_.Resize(fieldInfo.fixedFieldBytes, fieldInfo.varFieldCount);
```
rldoris: Should `Resize` be capitalized?

seanovens: I originally capitalized to distinguish it from `std::vector::resize`, but I suppose by that logic I should have also capitalized `FixedFieldRegion::size`, `FixedFieldRegion::data`, `FlatFieldArray::size`, and `FlatFieldArray::data`. I can change the name to `resize` to match the other "stdlib-like" methods.

docs/messagebody.md
seanovens: FYI I removed the non-`const` accessors in [a7afee4](https://github.com/novatel/novatel_edie/pull/241/commits/a7afee4dfd3b9d59ac950bef5e8c58019048b4ab)

riley-kinahan (approved):
Change look good, no more feedback on my end!

rldoris (approved):
Approved

PR #242, opened 2026-06-01
Add support for message object creation and mutation

riley-kinahan:
# PR Summary

## 1. Overview

This branch turns the Python message object graph from a **read‑only view over decoded data** into a **fully constructible and mutable API**.

Previously, `Field`/`Message` objects could only be produced by the decoder and inspected: `Field.__init__`/`Message.__init__` raised *"…initialization not implemented"*, there was no `__setattr__`, `FieldArray` exposed only `__getitem__`/`__len__`, and every `Header`/`RecieverStatus`/`MessageType` attribute was read‑only. All of that is now writable and constructible from Python.

### Messages

- **Construct by keyword:** `BESTPOS(latitude=45.5, longitude=-122.3)`. Every settable field is accepted as a keyword argument; omitted fields take type‑appropriate defaults.
- **Optional header:** `Message(header=..., **fields)`. A supplied `Header` is copied onto the message; `header=None` (the default) starts from a default header. Regardless of what the caller passes, `message_id` and `message_definition_crc` are taken from the message definition, so a message always identifies as itself.
- **Mutate after construction:** `m.latitude = 12.34`.
- **Encode/decode round‑trip:** a message built in Python can be encoded and decoded back to an equal message.
- **Validation:** unknown field names (as kwargs or via `setattr`) raise `AttributeError`; type mismatches raise `TypeError`.

### Fields & attribute assignment

Assignment is dispatched by field type:

| Field type | Accepts | Notes |
|---|---|---|
| `SIMPLE` | matching Python scalar | cast to the field's `DATA_TYPE`; non‑numeric / out‑of‑range rejected; `HEXBYTE`/`SATELLITEID` intentionally not settable |
| `ENUM` | `int` or `IntEnum` member | stored as int32; reads back as the bound `IntEnum`; values with no enumerator read back as `EDIE_UNKNOWN` |
| `STRING` | `str` | |
| `FIXED_LENGTH_ARRAY` | `list` / (`str` / `bytes` if `!isCsv`) | zero‑padded up to the array length |
| `VARIABLE_LENGTH_ARRAY` | `list` / (`str` / `bytes` if `!isCsv`) | stored as given |
| `FIELD_ARRAY` | `FieldArray` or `list[Field]` | Underlying `Field` type must match |

- **Length fields are read‑only:** assigning `*_length` raises `AttributeError` (*"Length cannot be set directly…"*) — modify the corresponding array instead.
- **Defaults:** a freshly constructed message zero‑initializes scalars, empties variable arrays, zero‑pads fixed arrays, and leaves enums at `0` (which reads back as `EDIE_UNKNOWN`).

### FieldArray

- **Construct explicitly or implicitly:** `points=FieldArray([Point(x=1, y=2), …])`, or just `points=[Point(...), …]`.
- **Index get/set:** `arr[i]` returns a live element wrapper; `arr[i] = Point(...)` replaces it (element type must match).
- **Nested field arrays** are supported to arbitrary depth.
- **Independence:** reassigning or dropping a message's array leaves any previously‑held element/array objects valid and unchanged.
- **Limits:** exceeding the definition's `array_length` raises `ValueError`; assigning a non‑`Field` element raises `TypeError`.
- **Not yet implemented:** `del arr[i]` and `arr.append(...)` (present as skipped tests).

### Standard (scalar) arrays

- `%s`‑conversion arrays behave as strings (fixed and variable alike) and accept `str`/`bytes`/`list`. 
- String-like arrays that do not represent "comma seperated values" accept string or byte arguments but are returned as lists of ints.
- Numeric arrays accept lists; strings/bytes are rejected for them.
- Fixed arrays read back zero‑padded to length; oversize values raise `ValueError`.

### Header, RecieverStatus, MessageType

- **`Header`** is constructible by keyword (all fields) and every attribute is read‑write. The `message_type`/`receiver_status` setters accept either the wrapper object or a raw int. `time_status` returns a `TIME_STATUS` for known values and a plain `int` for unknown ones.
- **`RecieverStatus`** is fully read‑write: construct from a raw `uint32` or from individual boolean flags (keyword‑only), read/write each flag as a property, plus a proper 2‑bit `version_bits` field typed as the new `RECEIVER_STATUS_VERSION` enum. Supports `==`.
- **`MessageType`** constructs from a raw `uint8` or from `is_response`/`format`/`sibling_id` (keyword‑only), exposes `raw_value`, and supports `==`.
- Equality on `RecieverStatus`/`MessageType` holds only against another instance of the same type with the same raw value.

### MessageDatabase

- **Constructor:** `MessageDatabase(file_path=None, message_family=None)` — a single signature replacing the previous two. Passing neither argument logs a one‑time warning recommending `message_family=`.
- **`get_field_type(msg_name, field, crc=None)`** returns the Python type for a `FieldArray` field, where `field` is a name or a list of names for nested field arrays and `crc` defaults to the latest. Bad message/CRC/field paths raise `KeyError`/`TypeError`/`ValueError`.

### Type stubs

- `FIELD_ARRAY` fields are now hinted as `FieldArray[<…>_Field]` (was `list[...]`), backed by a hand‑written generic `FieldArray[T]` stub.
- Each generated message class now has a real `__init__(self, header=None, <field>=..., …)` signature listing every settable field.

---

### Field storage & lifetime — `field_objects.hpp`, `implementations/field_objects.cpp`

- **Own‑or‑borrow storage** (`std::vector fields` + `nb::object parent` → `std::variant<unique_ptr<vector<FieldContainer>>, nb::object>`). Enforces the invariant that the data has a single owner.
- **`BaseField::ConstPtr` instead of raw `const BaseField*`** (through the field‑type lookup, `PyField`/`PyFieldArray::fieldDef`, and the DB map). `FieldContainer`s are now *minted on the Python side* (default construction and per‑field assignment), and each one must be handed a shared pointer to its definition. The decoder used to be the only producer of containers, so a raw pointer sufficed before.
- **`PyFieldArray` made a `class` with copy/move deleted.** It is always managed by pointer behind an `nb::object`; copying it would corrupt the cached‑wrapper / weakref bookkeeping.
- **Field‑array assignment copies rather than moves.** Profiling showed the copy is dwarfed by the per‑value conversions and C++/Python interop that dominate construction; a "move when the source owns its data" optimization yielded no top‑level detectable improvement, so it was deliberately left out.

### Construction & mutation entry points — `implementations/field_objects.cpp`, `bindings/field_objects.cpp`

- **Added `py_new` / `py_init` / `setattr` / `BuildDefaultFields` / `get_regular_array`, and bound `Field.__setattr__` and `FieldArray.__setitem__`.** This is the machinery that makes construction and mutation possible at all — the previous `__init__` was a throwing stub.

### Message database — `message_database.*`

- **Constructor unified into `(file_path=None, message_family=None)` with a no‑family warning** (*"Fix messageDB constructor"*). Lets a database be created empty *and* tagged with a family in one call; the warning nudges callers toward setting a family. Existing tests now pass `message_family="OEM"` to match.
- **`get_field_type` added.** Constructing a `FieldArray` requires the element's Python type, which callers previously had no way to obtain.
- **Field‑name alias gains a `_Field` suffix** (*"Fix field name aliasing"*). The alias must match the type name `AddFieldType` actually generates, which already ended in `_Field`.

### OEM header objects — `py_message_objects.*`

- **`RecieverStatus` / `MessageType` / `Header` made writable & constructible.** Required for building messages and headers from scratch in Python.
- **`version_bits` corrected** to a real 2‑bit field with a `RECEIVER_STATUS_VERSION` enum. The old getter (`(value >> 25) & 0x06000000`) masked bits that had already been shifted out and was effectively always wrong.
- **`reserved` flag now exposed; `__eq__` added.** Completeness for round‑tripping and value comparison.
- **`time_status` returns `TIME_STATUS | int`.** Tolerate raw values that have no enumerator instead of throwing.

### OEM messages & responses

- **`Message.__new__` / `__init__` rewritten** to default‑construct and then apply kwargs, force‑setting `message_id` / `message_definition_crc` from the definition. A constructed message must be internally consistent and identify as itself when encoded.
- **`PyResponse` and `PyEncode` now read through `fieldsPtr` / `owned_fields()`.** Mechanical consequence of removing the old `fields` member.

### Stub generation — `stubgen.py`, `common_stubgen_pattern.txt`, `CMakeLists.txt`

- **`FieldArray[T]` hints, per‑message `__init__` signatures, and a generic `FieldArray` stub injected via `--pattern-file`.** The stubs must advertise the now‑constructible API and the real `FieldArray` type so editors and type‑checkers see the new surface.

### Package exports

- **Re‑exported `MESSAGE_FORMAT` (common) and `MessageType` / `RecieverStatus` / `RECEIVER_STATUS_VERSION` (oem).** These types are now part of the public constructor interface.

---

## 4. Test coverage

New, mostly data‑driven coverage of the new API:

- **`test/test_message_api/`** (synthetic message definitions, independent of the shipped OEM DB):
  - `test_simple.py` (363 lines) — SIMPLE scalars for every `DATA_TYPE`: round‑trips across full integer ranges, float precision / extremes / ±inf, return types, zero defaults, and validation (non‑numeric, out‑of‑range, unsettable `HEXBYTE`/`SATELLITEID`).
  - `test_enum.py` (214 lines) — ENUM fields over two enum layouts: assignment by int and member, retrieval as `IntEnum`, the `EDIE_UNKNOWN` fallback, defaults, non‑integer rejection.
  - `test_standard_array.py` (242 lines) — `%s` string arrays, `%P` byte buffers, fixed/variable numeric arrays, zero‑padding, length‑field and oversize rejections.
  - `test_field_array.py` (287 lines) — explicit/implicit construction, nesting, reassignment and element assignment with original‑object independence, empty/dropped arrays, oversize and wrong‑type rejection; `del`/`append` marked `skip('Not implemented')`.
- **`test/test_oem_header.py`** (305 lines) — `Header`/`RecieverStatus`/`MessageType` construction & setattr, `TIME_STATUS` enum/int fallback, per‑bit flag mapping, `version_bits`, equality.
- **`test/test_oem_message_smoke.py`** (177 lines) — construction/mutation/encode‑decode against real OEM messages (`BESTPOS`, `BESTPOS_B1F6`, `RANGE`, `PASSCOM1`), header‑override behavior, and general validation.
- **`test_message_database.py`** — updated to the new constructor signature in seven places.

---

## 5. Performance Report

To analyze the performance I compared constructing EDIE message objects against constructing classes with identical fields directly in python. 

### BESTPOS (flat scalar/enum message, 17 fields)

| Implementation        | msg/s     |
| --------------------- | --------- |
| pure-Python `PyBestpos` | 849,820   |
| edie `BESTPOS`          | 171,338   |

### RANGE (field array, 325 obs each)

| Implementation                  | msg/s    | obs/s     |
| ------------------------------- | -------- | --------- |
| pure-Python `PyRange` + `PyObs` | 2,076.1  | 674,731   |
| edie `RANGE`                    | 914.7    | 297,265   |
| edie columnar `RANGE` (shown below)   | 3,122.4    | 1,014,787   |

Construction is definitely a bit slow but within the realm of what's expected for constructing complex objects in python. Based on my profiling I think a ~1.3x improvement is possible with some minor optimizations. @seanovens refactor work will probably make it a little faster as well but I'm not fully sure on that.

I also experimented with a alternative columnar based scheme that yields about 3x better performance that can be added in a future update:

```
        m = RANGE(obs=FieldArray(
                RANGE_obs_Field,
                sv_prn=list(range(MAX_OBS)),
                sv_freq=[0 for i in range(MAX_OBS)],
                psr=[20000000.0 + i for i in range(MAX_OBS)],
                sd_psr=[0.5 for i in range(MAX_OBS)],
                adr=[12345.6 + i for i in range(MAX_OBS)],
                sd_adr=[0.01 for i in range(MAX_OBS)],
                dop=[1500.0 + i for i in range(MAX_OBS)],
                C_No=[45.0 for i in range(MAX_OBS)],
                lock_time=[100.0 for i in range(MAX_OBS)],
                c_status=[0 for i in range(MAX_OBS)],
            )
        )
```

---

## 6. Notes & future actions

- **String‑array heuristic** still keys on `conversion == "%s"`; a TODO marks switching to an `isString` check, left as‑is to avoid breaking user code.
- **`BaseField::ConstPtr` plumbing** could be reverted to raw pointers if `FieldContainer` is changed to hold a raw definition pointer. It only carries shared ownership because `FieldContainer` does; the owning `MessageDatabase` is already guaranteed alive for any field's lifetime (Python holds a `PyMessageDatabase::ConstPtr parentDb`, and C++ callers could be expected to keep the database alive while using messages decoded against it).
- **Field Construction Optimization**: Add an alternative columnar based constructor and implement in-place construction for provided values rather than current "default then set" pattern.


python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -190,22 +315,76 @@ void py_oem::init_header_objects(nb::module_& m)
...
+                    return nb::cast(static_cast<TIME_STATUS>(self.uiTimeStatus));
+                }
+                catch (const std::exception&)
+                {
+                    return nb::cast(self.uiTimeStatus);
+                }
+            },
+            [](py_oem::PyHeader& self, uint32_t value) { self.uiTimeStatus = value; }, nb::sig("def time_status(self) -> TIME_STATUS | int"),
+            "The quality of the GPS reference time.")
+        .def_rw("week", &py_oem::PyHeader::usWeek, "GPS reference wekk number.")
```
seanovens:
```suggestion
        .def_rw("week", &py_oem::PyHeader::usWeek, "GPS reference week number.")
```

python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -190,22 +315,76 @@ void py_oem::init_header_objects(nb::module_& m)
...
-        .def_prop_ro(
-            "time_status", [](const py_oem::PyHeader& self) { return TIME_STATUS(self.uiTimeStatus); }, "The quality of the GPS reference time.")
-        .def_ro("week", &py_oem::PyHeader::usWeek, "GPS reference wekk number.")
-        .def_ro("milliseconds", &py_oem::PyHeader::dMilliseconds, "Milliseconds from the beginning of the GPS reference week.")
-        .def_prop_ro("receiver_status", &py_oem::PyHeader::GetRecieverStatus,
-                     "32-bits representing the status of various hardware and software components of the receiver.")
-        .def_ro("message_definition_crc", &py_oem::PyHeader::uiMessageDefinitionCrc,
+        .def(
+            "__init__",
+            [](py_oem::PyHeader* self, uint16_t message_id, std::variant<py_oem::PyMessageTypeField, uint8_t> message_type, uint32_t port_address,
```
seanovens: Given that `PyMessageTypeField` is implicitly constructible from `uint8_t`, would it make sense for `message_type` to simply be a `PyMessageTypeField` to avoid the variant handling? (See also the `message_type` binding below)

python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -190,22 +315,76 @@ void py_oem::init_header_objects(nb::module_& m)
...
+                self->usSequence = sequence;
+                self->ucIdleTime = idle_time;
+                self->uiTimeStatus = time_status;
+                self->usWeek = week;
+                self->dMilliseconds = milliseconds;
+                self->uiReceiverStatus = std::holds_alternative<py_oem::PyRecieverStatus>(receiver_status)
+                                             ? std::get<py_oem::PyRecieverStatus>(receiver_status).value
+                                             : std::get<uint32_t>(receiver_status);
+                self->uiMessageDefinitionCrc = message_definition_crc;
+                self->usReceiverSwVersion = receiver_sw_version;
```
seanovens, line 360: Also need to set `self->format` here, otherwise there can be issues encoding custom messages. Right now, if you do `oem_msgs.BESTPOS().to_binary()`, you'll get a binary BESTPOS message with 72 in the _header_ length field. This is because when a new header is constructed its `format` defaults to 0, which isn't a valid value for a `HEADER_FORMAT`. This causes the header encoder to encode the header as BINARY, while the body encoder treats the header as SHORT_BINARY and puts the message length in the wrong spot (side note: the encoder should probably fall back to the same format for the header and the body in cases like this).

Maybe it also makes sense to make `PyHeader::format` writable? Otherwise I don't think it is possible to encode a custom message as SHORT_BINARY/SHORT_ASCII, for example. (On the other hand, this could make for a bit of a confusing interface where the `PyEncodableField::to_x()` methods behave differently depending on the header's `format`.)

python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -190,22 +315,76 @@ void py_oem::init_header_objects(nb::module_& m)
...
+            [](const py_oem::PyHeader& self) -> nb::object {
+                try
+                {
+                    return nb::cast(static_cast<TIME_STATUS>(self.uiTimeStatus));
+                }
+                catch (const std::exception&)
+                {
+                    return nb::cast(self.uiTimeStatus);
+                }
+            },
```
seanovens, line 385: Would it make sense to define an `IsValidTimeStatus` helper method rather than throwing/catching? My only concern is that default-constructed headers have time status 0, i.e. not a valid `TIME_STATUS`, so the exception cost would be paid for every `time_status` access on a default-constructed message. Alternatively, perhaps it would make sense for the default time status to be some valid value, e.g. 20.

python/src/py_oem/bindings/py_message_objects.cpp
```diff
@@ -190,22 +315,76 @@ void py_oem::init_header_objects(nb::module_& m)
...
-        .def_ro("week", &py_oem::PyHeader::usWeek, "GPS reference wekk number.")
-        .def_ro("milliseconds", &py_oem::PyHeader::dMilliseconds, "Milliseconds from the beginning of the GPS reference week.")
-        .def_prop_ro("receiver_status", &py_oem::PyHeader::GetRecieverStatus,
-                     "32-bits representing the status of various hardware and software components of the receiver.")
-        .def_ro("message_definition_crc", &py_oem::PyHeader::uiMessageDefinitionCrc,
+        .def(
+            "__init__",
+            [](py_oem::PyHeader* self, uint16_t message_id, std::variant<py_oem::PyMessageTypeField, uint8_t> message_type, uint32_t port_address,
+               uint16_t length, uint16_t sequence, uint8_t idle_time, uint32_t time_status, uint16_t week, double milliseconds,
+               std::variant<py_oem::PyRecieverStatus, uint32_t> receiver_status, uint32_t message_definition_crc, uint16_t receiver_sw_version) {
```
seanovens: Should `message_id` and `message_definition_crc` be read only since they are overwritten in the message constructor anyways?

python/src/py_common/implementations/field_objects.cpp
```diff
@@ -61,21 +159,23 @@ PYCOMMON_EXPORT nb::object py_common::PyField::convert_field(FieldContainer& fie
...
+            nb::object pyArr = nb::inst_alloc(nb::type<PyFieldArray>());
+            new (nb::inst_ptr<PyFieldArray>(pyArr)) PyFieldArray(message_field, field.fieldDef, parentDb, nb::cast(this, nb::rv_policy::none));
+            nb::inst_mark_ready(pyArr);
             cachedArrays_[fieldIdx] = nb::weakref(pyArr);
             return pyArr;
         }
         else
         {
             // Handle Fixed or Variable-Length Arrays
+            // TODO: This check should be for 'isString' but changing it could break a lot of user code
```
seanovens: Would the only difference be the possibility of the conversion string "%S"?

riley-kinahan: In practice I think that's right, the FIELD_TYPE::STRING seems to dictate a %s conversion string. 

I added that comment without much investigation but I think the change is safe to make. I see documentation about `%S` being used for a special output formatting but I don't think it exists in any of our logs currently.

python/src/py_oem/bindings/py_message_objects.cpp
riley-kinahan: On the python side the `TIME_STATUS` is defaulted to `TIME_STATUS.UNKNOWN` when constructing a header, but I think you're suggestion is cleaner anyways.

python/src/py_oem/bindings/py_message_objects.cpp
riley-kinahan: Yes, I think this better protects against misuse. I'll also make length read only since it doesn't have any logical meaning for a constructed message.

python/src/py_oem/bindings/py_message_objects.cpp
riley-kinahan: Good thought, it looks like nanobind allows exposing an implicit conversion which will make this behavior more universal if we `MessageType` ever shows up as a field elsewhere.

python/src/py_oem/bindings/py_message_objects.cpp
riley-kinahan: Thanks for catching this bug!

For now I've explicitly set the header format to `UNKNOWN` and updated the body encoder to default to long header format. Confirmed with a new test that encodes a `BESTPOS` message, decodes it, and compares the header length against the actual length of the encoded data. This fails against the previous code.

My long term suggestion for a fix is that `PyHeader::format` be removed in favor of adding a "header_type" field to message definitions as the source of truth.

seanovens: Supporting an optional "short" parameter in `to_binary/to_ascii` for `oem::PyMessage` could also make sense. That would match the C++ interface a bit more closely, and it might make sense to allow encoding any message in either its short/normal form without changing its definition. Maybe @rldoris has an opinion?

rldoris: Depends on how its implemented. We used to support generic "generate short/full" messages, but decided it wasn't worth the pain of people generating full headers on wrong messages messages and the confusion it would cause. 

I header type is a field we should add to the JSON and have it default to full headers if it's missing. It isn't worth delaying this MR over though so I say we just press on.

python/src/py_common/bindings/message_database.cpp
```diff
@@ -277,9 +282,19 @@ void py_common::init_common_message_database(nb::module_& m)
...
 
     nb::class_<py_common::PyMessageDatabase>(m, "MessageDatabase", nb::type_slots(db_slots))
-        .def(nb::new_([]() { return py_common::PyMessageDatabase::Create(); }))
-        .def(nb::new_([](std::filesystem::path& file_path) { return py_common::PyMessageDatabase::Create(std::move(*LoadJsonDbFile(file_path))); }),
-             "file_path"_a)
+        .def(nb::new_([](std::optional<std::filesystem::path> file_path, std::optional<std::string> message_family) {
+                 nb::object wrapper = file_path.has_value() ? py_common::PyMessageDatabase::Create(std::move(*LoadJsonDbFile(*file_path)))
+                                                            : py_common::PyMessageDatabase::Create();
+                 if (message_family.has_value()) { nb::cast<py_common::PyMessageDatabase*>(wrapper)->SetMessageFamily(*message_family); }
+                 else if (!file_path.has_value())
```
seanovens: Is there a reason the python-side missing message family warning is suppressed when a file path is provided?

python/src/py_common/implementations/field_objects.cpp
```diff
@@ -230,4 +341,218 @@ PYCOMMON_EXPORT nb::object PyFieldArray::getitem(ssize_t signedIndex) const
...
+    }
+
+    for (auto it = values.begin(); it != values.end(); it++)
+    {
+        if (!nb::isinstance<PyField>(nb::handle(*it))) { throw nb::type_error("Only Fields can appear within a FieldArray!"); }
+    }
+    PyField* candidate = nb::inst_ptr<PyField>(nb::handle(*values.begin()));
+    fieldDef = candidate->fieldDef;
+    parentDb = candidate->parentDb;
+    const auto* arrayDef = static_cast<const ArrayField*>(fieldDef.get());
```
seanovens: Maybe worth doing dynamic cast + nullptr check here to avoid hard crash e.g. if someone passes a message object here

seanovens (commented):
Ok everything else looks good to me upon second viewing!

python/src/py_common/bindings/message_database.cpp
riley-kinahan: Nope, switching to a check for an empty message family is more robust.

python/src/py_common/implementations/field_objects.cpp
riley-kinahan: Good catch, added the check and a test for it

rldoris:
Haven't had a ton of time to dig into this nearly as much as I should, but on the surface everything looks good (unit tests passing, benchmarks not complaining, and no major interface changes) so I'm reasonably comfortable approving.

rldoris (approved):

PR #245, opened 2026-06-02
Drop nlohman json

riley-kinahan:
# Drop nlohmann-json in favour of simdjson

Replaces the nlohmann/json dependency with simdjson across the JSON message decoder and the JSON message-database reader. Decoding was meant to have been swapped already but a couple places hadn't been fully updated. 

Note: As cleanup the unused `constellation` from `MetaDataBase` is also being removed.

## Decoder changes

OEM JSON decode functions migrated from `nlohmann::json` to `simdjson::dom::element`. Previously this code was broken.

- **`oem/src/message_decoder.cpp`** — All `jsonFieldMap` lambdas now take `simdjson::dom::element` instead of `const json&`. Reads go through `.get(out)` with an explicit `throw std::runtime_error("Failed to decode JSON field '...'")` on failure, replacing nlohmann's `.get<T>()` (which threw internally). 
- **`common/message_decoder.hpp`** — `PushElement<T>` now throws on a failed `.get()` instead of silently doing nothing. `SimpleJsonMapEntry<T>` and related signatures updated to `simdjson::dom::element`.
- **`common/src/message_decoder.cpp`** — Added a `try`/`catch (const std::runtime_error&)` around the `FIELD_TYPE::SIMPLE` decode that converts a decode exception into `STATUS::MALFORMED_INPUT` (with a `CRITICAL` log) rather than letting it escape. This matches ascii decoding.

## Reader changes (JSON DB reader)

Database loading now uses simdjson.

- **New helpers** (anonymous namespace) for accessing data out of simdjson: `Member` (required key, throws if absent), `AsStringView` / `AsString`, `AsStringOrEmpty` (null → `""`), `AsUint`, and `StringOr` (optional key with default).
- **Explicit parse functions** — the `from_json(const json&, T&)` overloads are replaced with named `ParseEnumDataType` / `ParseSimpleDataType` / `ParseBaseField` / `ParseEnumField` / `ParseArrayField` / `ParseFieldArrayField` / `ParseMessageDefinition` / `ParseEnumDefinition` / `ParseDbMetadata` / `ParseFields` / `ParseEnumerators`, each taking a `simdjson::dom::element`. These functions are all essentially the same as the old ones but just use the new helpers instead of nholman_json's accessors.

## Build / tooling / dependency changes

- Drop all references to nlohman_json from conan and cmake.
- **`.github/workflows/build_cpp.yml`** — add `--gcov-ignore-parse-errors=negative_hits.warn_once_per_file` to the `gcovr` coverage steps to work around a gcc bug producing negative hit counts. simdjson trips this issue.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -49,4 +49,5 @@ HeaderFilterRegex: 'src/.*|examples/.*'
 
 WarningsAsErrors: '*'
 
-ExtraArgs: - -Wno-unknown-pragmas # TODO: remove this after switch to clang-tidy-18
+# Explictly allow exceptions, otherwise MSVC with clang-tidy runs into issues
+ExtraArgs: [-Wno-unknown-pragmas, -Xclang, -fcxx-exceptions]
--- a/include/novatel_edie/decoders/common/message_decoder.hpp
+++ b/include/novatel_edie/decoders/common/message_decoder.hpp
@@ -241,15 +241,12 @@ class MessageDecoderBase
         static_assert(!std::is_same_v<IntermediateType, void>, "Unsupported type for PushElement");
 
         IntermediateType intermediateValue;
-        if (clJsonField_.get(intermediateValue) == simdjson::SUCCESS)
+        if (clJsonField_.get(intermediateValue) != simdjson::SUCCESS)
         {
-            T value = static_cast<T>(intermediateValue);
-            vIntermediate_.emplace_back(value, std::move(pstMessageDataType_));
+            throw std::runtime_error("Failed to decode JSON field '" + pstMessageDataType_->name + "'");
         }
-        else
-        {
-            // Handle error (e.g., log a warning or throw an exception)
-        }
+        T value = static_cast<T>(intermediateValue);
+        vIntermediate_.emplace_back(value, std::move(pstMessageDataType_));
     }
 
     // -------------------------------------------------------------------------------------------------------
--- a/src/decoders/common/src/json_db_reader.cpp
+++ b/src/decoders/common/src/json_db_reader.cpp
@@ -193,10 +193,6 @@ void ParseMessageDefinition(element j_, MessageDefinition& md_)
     md_.description = AsStringOrEmpty(Member(j_, "description"));
     md_.latestMessageCrc = std::stoul(AsString(Member(j_, "latestMsgDefCrc")));
 
-    element messageStyle;
-    md_.messageStyle =
-        j_["messageStyle"].get(messageStyle) == simdjson::SUCCESS && !messageStyle.is_null() ? AsString(messageStyle) : "OEM4_MESSAGE_STYLE";
-
     object fields;
     if (Member(j_, "fields").get(fields) != simdjson::SUCCESS) { throw std::runtime_error("Expected 'fields' to be a JSON object"); }
     for (auto field : fields)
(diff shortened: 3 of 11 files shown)


.clang-tidy
```diff
@@ -49,4 +49,5 @@ HeaderFilterRegex: 'src/.*|examples/.*'
 
 WarningsAsErrors: '*'
 
-ExtraArgs: - -Wno-unknown-pragmas # TODO: remove this after switch to clang-tidy-18
+# Explictly allow exceptions, otherwise MSVC with clang-tidy runs into issues
+ExtraArgs: [-Wno-unknown-pragmas, -Xclang, -fcxx-exceptions]
```
riley-kinahan, line 53: Not related to these changes, just a small cleanup to improve developer experience on windows

include/novatel_edie/decoders/common/message_decoder.hpp
riley-kinahan, line 230: Now raises an exception on failure. All calls propagate to one catch block

```
            try
            {
                DecodeJsonField(field, clField, vIntermediateFormat_);
            }
            catch (const std::runtime_error&)
            {
                SPDLOG_LOGGER_CRITICAL(pclMyLogger, "DecodeJsonField(): Exception when decoding field. Malformed Input\n");
                return STATUS::MALFORMED_INPUT;
            }
            break;
```

src/decoders/common/src/message_decoder.cpp
```diff
@@ -853,7 +853,17 @@ STATUS MessageDecoderBase::DecodeJson(const std::vector<BaseField::Ptr>& vMsgDef
...
+            try
+            {
+                DecodeJsonField(field, clField, vIntermediateFormat_);
+            }
+            catch (const std::runtime_error&)
+            {
+                SPDLOG_LOGGER_CRITICAL(pclMyLogger, "DecodeJsonField(): Exception when decoding field. Malformed Input\n");
+                return STATUS::MALFORMED_INPUT;
+            }
+            break;
```
riley-kinahan, line 866: Mirrors behavior of ascii decoder

src/decoders/common/src/json_db_reader.cpp
```diff
@@ -26,208 +26,319 @@
...
+    {
+        EnumDataType enumDataType;
+        ParseEnumDataType(enumerator, enumDataType);
+        vEnumerators_.emplace_back(std::move(enumDataType));
+    }
 }
 
 //-----------------------------------------------------------------------
-std::vector<MessageDefinition::ConstPtr> ProcessMessageDefinitions(const json& jArray)
+std::vector<MessageDefinition::ConstPtr> ProcessMessageDefinitions(element jRoot_)
```
rldoris, line 285: Should we keep `jRoot_` as a const or does the `.get()` function not like it?

riley-kinahan: `simdjson::dom::element` is a read only view similar to `std::string_view` so its effectively the same as a const reference. Happy to update if you think the explicitness provides value, but we aren't losing any safety as is.

rldoris: The compiler can make some optimizations knowing it's `const`, but it's having the built in constness is good enough.

rldoris (approved):

rldoris (approved):

PR #246, opened 2026-06-02
Range decompressor bug fixes

seanovens:
This PR contains several bug fixes for RANGECMP4 decompression and RANGECMP4/5 lock time tracking.

### RANGECMP4 decompression

- Previously, when a differential block was skipped due to missing reference data, the decompressor moved on to the next signal without decoding the differential block. But the bit offset/buffer pointer still pointed at the "discarded" differential block, meaning that the remaining signals in the log could not be decoded. To fix this, I've added code that advances the buffer pointer and bit offset to skip over the discarded differential block.
- Missing reference data also used to throw/catch an exception because the map lookup was done using `at()`; I've changed this to  `find()` + check, which should be significantly faster when reference data is frequently missing.

### RANGECMP4/5 lock time tracking

- Previous code expected the accumulated lock time not to exceed `2 * lockTime[ucLockTimeBits_]`. Since `lockTime[0] = 0.0`, this check would always fail when the lock time bits were 0 (causing a bunch of "Expected a bit change much sooner" warnings to be emitted).
- This check would continue to be performed when the lock time index had reached its maximum value of 15, meaning that the warning would be emitted every 262144ms as long as lock was maintained.
- I've changed this so that the check is only performed when in absolute mode and when the lock time index is < 15. The condition now checks that the accumulated lock time is no larger than `lockTime[ucLockTimeBits_ + 1]` (since table entries are consecutive powers of 2 this is essentially the same as before except when `ucLockTimeBits_` is 0 or 15)

### DIFF
--- a/include/novatel_edie/decoders/oem/rangecmp/common.hpp
+++ b/include/novatel_edie/decoders/oem/rangecmp/common.hpp
@@ -330,6 +330,14 @@ constexpr std::array<uint32_t, 2> RBLK_DOPPLER_BITS = {26, 14};
 constexpr std::array<uint32_t, 2> RBLK_DOPPLER_SIGNEXT_MASK = {0xFC000000, 0xFFFFC000};
 constexpr std::array<int64_t, 2> RBLK_INVALID_PSR = {137438953471, -524288};
 
+constexpr uint32_t SIG_DBLK_PRIMARY_BITS = SIG_BLK_PARITY_FLAG_BITS + SIG_BLK_HALF_CYCLE_BITS + SIG_BLK_CNO_BITS + SIG_BLK_LOCK_TIME_BITS +
+                                           SIG_BLK_PSR_STDDEV_BITS + SIG_BLK_ADR_STDDEV_BITS + SIG_DBLK_PSR_BITS + SIG_DBLK_PHR_BITS +
+                                           DBLK_DOPPLER_BITS[0];
+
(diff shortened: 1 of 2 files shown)


src/decoders/oem/src/rangecmp/range_decompressor.cpp
riley-kinahan: Could you clean up some of our `reinterpret_cast` usage while you are at it?

From my understanding the usage of `reinterpret_cast` here is undefined behavior because `*pucData_` isn't actually an integer. I suspect this is unlikely to be an issue in any actual compiler but I think the blessed way to do this is by using std::memcpy like in the `std::bit_cast` implementation advice:
https://en.cppreference.com/cpp/numeric/bit_cast#:~:text=Possible%20implementation

include/novatel_edie/decoders/oem/rangecmp/common.hpp
riley-kinahan, line 151: This has the same `reinterpret_cast` issue as mentioned below but I'm also unsure if we can safely access 8 bytes here or if it is possibly past the end of the buffer.

rldoris: I'm going to change the other reinterpret_cast, but leave this one alone for now. I suspect a dedicated review & fix of our uses of reinterpret_cast would be better than trying to cram it into a bug fix.

PR #250, opened 2026-06-25
Remove all unsafe reinterpret casts

seanovens:
## Summary

- Replaces all calls to `reinterpret_cast<T*>(...)` where `T` $\not\in$ {`char`, `unsigned char`, `std::byte`} with a safe alternative
- Removes `reinterpret_cast` from C++ field access instructions

## Explanation

The C++ standard only relaxes strict type aliasing for `char`, `unsigned char`, and `std::byte`, so the compiler is allowed to assume that pointers to any other types cannot overlap. Therefore, dereferencing a pointer like `reinterpret_cast<T*>(ptr)` has **undefined behaviour** if `T` is not a `char`, `unsigned char`, or `std::byte`, even when the alignment of `T` is just as strict as the alignment of `ptr`'s type.

Note: as far as I understand, in almost all cases `int8_t`/`uint8_t` are simply aliases for `char`/`unsigned char`, respectively, so dereferencing a pointer obtained through `reinterpret_cast<int8_t*/uint8_t*>(...)` is generally fine. However, in theory it is possible that these types are defined as some distinct extended integer types different from `char`/`unsigned char`, so I decided to remove all reinterpret casts to `int8_t*`/`uint8_t*` as well.

## Additional changes

- Changed buffer types for some encoder/commander functions from `unsigned char**` to `unsigned char* const*` to better reflect function behaviour
- Added `BufferType` template argument to `CopyAllToBuffer` to match other `Copy...` functions and prevent needing to keep switching from `unsigned char**` to `char**`

## Performance impact

Binary decoding/encoding performance seems to be negatively affected by these changes on the runner. I have reproduced this slowdown locally (though the slowdown is much smaller on my machine i.e. <5% for binary decoding). The slowdown only seems to happen when compiling with `g++` on Linux - I observe no significant difference in performance (on my local machine) when compiling with MSVC on Windows or `clang++` on Linux. Interestingly, with Linux `g++` performance was _significantly_ improved (from ~40% slowdown to ~25%) by moving variable initialization/memcpy into a function (`LoadValueFromBuffer`) rather than initializing a local variable and memcpying directly inline.

Note that this performance impact is irrelevant for the message data structure refactor (#241), since binary decoding/encoding simply involves copying raw byte regions in the updated code.

### DIFF
--- a/include/novatel_edie/decoders/common/encoder.hpp
+++ b/include/novatel_edie/decoders/common/encoder.hpp
@@ -67,20 +67,19 @@ template <typename T> inline std::chars_format FloatingPointFormat(const FieldCo
 }
 
 // -------------------------------------------------------------------------------------------------------
-template <typename T> [[nodiscard]] bool WriteIntToBuffer(unsigned char** ppucBuffer_, uint32_t& uiBytesLeft_, T&& arg)
+template <typename T> [[nodiscard]] bool WriteIntToBuffer(char** ppcBuffer_, uint32_t& uiBytesLeft_, T&& arg)
 {
     static_assert(std::is_integral_v<std::decay_t<T>>, "Argument must be integral type.");
-    auto* ppcBuffer = reinterpret_cast<char*>(*ppucBuffer_);
-    auto [end, ec] = std::to_chars(ppcBuffer, ppcBuffer + uiBytesLeft_, arg);
+    auto [end, ec] = std::to_chars(*ppcBuffer_, *ppcBuffer_ + uiBytesLeft_, arg);
     if (ec != std::errc{}) { return false; }
-    const auto written = static_cast<uint32_t>(end - ppcBuffer);
-    *ppucBuffer_ = reinterpret_cast<unsigned char*>(end);
+    const auto written = static_cast<uint32_t>(end - *ppcBuffer_);
+    *ppcBuffer_ = end;
     uiBytesLeft_ -= written;
     return true;
 }
 
 // -------------------------------------------------------------------------------------------------------
-template <typename T> [[nodiscard]] bool WriteHexToBuffer(unsigned char** ppucBuffer_, uint32_t& uiBytesLeft_, T&& arg, uint32_t width)
+template <typename T> [[nodiscard]] bool WriteHexToBuffer(char** ppcBuffer_, uint32_t& uiBytesLeft_, T&& arg, uint32_t width)
 {
     static_assert(std::is_integral_v<std::decay_t<T>>, "Argument must be integral type.");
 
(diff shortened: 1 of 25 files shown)


riley-kinahan:
@seanovens How did you get the slowdown numbers for the gcc Linux case, was it reproducible?

FWIW x86-64 GCC-13.3 (used in the pipelines) seems to optimize simple calls to your `LoadValueFromBuffer` function:

image

seanovens:
@riley-kinahan I've just been running the `DecodeBinaryLog` benchmark locally through wsl. It does seem to be a bit slower (by around 5% or so), but not as significant as the slowdown in the pipeline benchmark

include/novatel_edie/decoders/common/encoder.hpp
riley-kinahan: Now that all the other higher level `ToBuffer` functions use a generic `BufferType` I think it makes sense to change it here too

include/novatel_edie/decoders/common/encoder.hpp
```diff
@@ -207,33 +207,43 @@ template <typename T, template <typename> class Template> struct is_specializati
 template <typename T, template <typename> class Template> inline constexpr bool is_specialization_of_v = is_specialization_of<T, Template>::value;
 
 // -------------------------------------------------------------------------------------------------------
-template <typename... Args> [[nodiscard]] bool CopyAllToBuffer(char** ppucBuffer_, uint32_t& uiBytesLeft_, Args&&... args_)
+template <typename BufferType, typename... Args> [[nodiscard]] bool CopyAllToBuffer(BufferType* buffer_, uint32_t& uiBytesLeft_, Args&&... args_)
 {
+    static_assert(std::is_same_v<BufferType, char*> || std::is_same_v<BufferType, unsigned char*>, "BufferType must be char* or unsigned char*.");
```
riley-kinahan: This is a good check to have. Worth copy pasting it to the other `ToBuffer` functions (maybe as a macro).

include/novatel_edie/decoders/common/encoder.hpp
```diff
@@ -243,24 +253,24 @@ template <typename BufferType, typename... Args>
 // -------------------------------------------------------------------------------------------------------
 template <typename T> std::function<bool(const FieldContainer&, char**, uint32_t&, const MessageDatabase&)> BasicMapEntry(const char* pcF_)
 {
-    return [pcF_](const FieldContainer& fc_, char** ppcOutBuf_, uint32_t& uiBytesLeft_, [[maybe_unused]] const MessageDatabase& pclMsgDb_) {
-        return WriteFormattedToBuffer(ppcOutBuf_, uiBytesLeft_, pcF_, std::get<T>(fc_.fieldValue));
+    return [pcF_](const FieldContainer& fc_, char** ppucOutBuf_, uint32_t& uiBytesLeft_, [[maybe_unused]] const MessageDatabase& pclMsgDb_) {
+        return WriteFormattedToBuffer(ppucOutBuf_, uiBytesLeft_, pcF_, std::get<T>(fc_.fieldValue));
```
riley-kinahan: Confused about this rename, looks like this is still a signed char buffer

src/decoders/oem/src/rangecmp/range_decompressor.cpp
```diff
@@ -445,7 +445,8 @@ void RangeDecompressor::RangeCmp2ToRange(const rangecmp2::RangeCmp& stRangeCmpMe
         for (uint8_t ucSignalBlockIndex = 0; ucSignalBlockIndex < ucSignalBlockCount; ucSignalBlockIndex++)
         {
             // Decompress the signal block
-            const auto& stSigBlock = reinterpret_cast<const SignalBlock&>(stRangeCmpMessage_.aucRangeData[uiRangeDataBytesDecompressed]);
+            SignalBlock stSigBlock;
+            std::memcpy(&stSigBlock, &stRangeCmpMessage_.aucRangeData[uiRangeDataBytesDecompressed], sizeof(SignalBlock));
```
riley-kinahan: Is there a reason the helper can't be used here?

src/decoders/oem/src/encoder.cpp
```diff
@@ -591,11 +602,15 @@ Encoder::EncodeBody(unsigned char** ppucBuffer_, uint32_t uiBufferSize_, const s
...
+            auto stHeader = LoadValueFromBuffer<Oem4BinaryHeader>(stMessageData_.pucMessageHeader);
+            stHeader.usLength = static_cast<uint16_t>(pucTempBuffer - *ppucBuffer_);
+            std::memcpy(stMessageData_.pucMessageHeader, &stHeader, sizeof(Oem4BinaryHeader));
         }
         else
         {
-            reinterpret_cast<Oem4BinaryShortHeader*>(stMessageData_.pucMessageHeader)->ucLength = static_cast<uint8_t>(pucTempBuffer - *ppucBuffer_);
+            auto stShortHeader = LoadValueFromBuffer<Oem4BinaryShortHeader>(stMessageData_.pucMessageHeader);
+            stShortHeader.ucLength = static_cast<uint8_t>(pucTempBuffer - *ppucBuffer_);
+            std::memcpy(stMessageData_.pucMessageHeader, &stShortHeader, sizeof(Oem4BinaryShortHeader));
```
riley-kinahan: Looks like GCC doesn't optimize this, it actually copies the data onto the stack, modifies it and copies it back in:

Image

Clang recognizes that it can just store the data directly:

Image

This more explicit code produces consistent minimal assembly across both compilers, let me know if updating it fixes the Linux gcc slowdown:

Image

riley-kinahan (changes requested):
That's all from me, nice job cleaning all this up!

include/novatel_edie/decoders/common/encoder.hpp
seanovens: Done 👍

include/novatel_edie/decoders/common/encoder.hpp
seanovens: Done 👍

include/novatel_edie/decoders/common/encoder.hpp
seanovens: Oops - I was flip-flopping on whether to just change everything to `unsigned char*` but decided against it. Somehow missed reverting this rename

On another note, I just realized this function isn't actually used anywhere so I decided to just remove it

src/decoders/oem/src/rangecmp/range_decompressor.cpp
seanovens: Thanks, missed this one

src/decoders/oem/src/encoder.cpp
seanovens: Great find, that appears to close the gap for the binary encoding performance at least 🥳

riley-kinahan (approved):

PR #251, opened 2026-06-30
FixedBuffer: Move the capacity out of the template to a member variable

rldoris:
Need to make a framer with larger capacity as the default 256 isn't large enough. Move the capacity/N out of the template and info a class member.

### DIFF
--- a/include/novatel_edie/common/fixed_buffer.hpp
+++ b/include/novatel_edie/common/fixed_buffer.hpp
@@ -35,6 +35,8 @@
 #include <type_traits>
 
 #include "novatel_edie/common/crc.hpp"
+//!< Recommended frame buffer size. Should work for most formats and message sizes. If the buffer is too small, the framer will return BUFFER_FULL.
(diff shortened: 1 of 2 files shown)


include/novatel_edie/common/fixed_buffer.hpp
```diff
@@ -188,12 +192,13 @@ template <typename T, size_t N> class FixedBuffer
...
+    size_t N;
+    std::unique_ptr<T[]> buffer;
     size_t head = 0;
     size_t sz = 0;
 };
 
 namespace novatel::edie {
 //! \brief FixedBuffer specialization for unsigned char with 256KB capacity.
-using UCharFixedBuffer = FixedBuffer<unsigned char, (1 << 18)>;
+using UCharFixedBuffer = FixedBuffer<unsigned char>;
```
seanovens, line 203: I think this makes more sense anyways - the name `UCharFixedBuffer` doesn't really suggest a fixed max size

seanovens (approved):

PR #257, opened 2026-07-13
Switch to CMakeConfigDeps to fix VS 2026 build

riley-kinahan:
The conan maintainer clarified that the CMakeConfigDeps generator is primarily marked as experimental because it is new: https://github.com/conan-io/conan/issues/20148#issuecomment-4912499844

In light of this I think it makes sense to switch over so that users don't have to update their config manually if they encounter build errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,35 +80,6 @@ The build can then be started with:
 cmake --build {path_to_build_folder}
 ```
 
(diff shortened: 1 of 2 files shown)


rldoris (approved):
Fixed my VS2026 issues. Thanks!

rldoris (approved):

PR #258, opened 2026-08-05
Restore: Switch to CMakeConfigDeps to fix VS 2026 build

seanovens:
Sorry @riley-kinahan, looks like I accidentally reverted this change as part of #241.

Also, I noticed that using `CMakeConfigDeps` still requires the `global.conf` modification with older versions of Conan (< 2.25.0). Do you think it is worth changing the required Conan version to >=2.25.0? Or do you think Conan's error message is informative enough, i.e.

> `ConanException: CMakeConfigDeps is being used in conanfile, but the conf 'tools.cmake.cmakedeps:new' is not enabled. This is an incubatingfeature, please define it to enable CMakeConfigDeps usage`

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -8,7 +8,7 @@ from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, cmake_layout, CMakeToolchain, CMakeConfigDeps
 from conan.tools.files import copy, rmdir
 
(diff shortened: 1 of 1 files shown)


riley-kinahan:
Updating the required conan version makes sense to me

riley-kinahan (approved):

PR #259, opened 2026-08-12
Add header types to messages

riley-kinahan:
Previously we relied on the format of a decoded message to determine whether to encode it with a standard or short header. This meant that directly constructed message objects would be encoded with an incorrect header type unless metadata was deliberately constructed with this in mind.

This PR adds a "headerType" field to the database which acts as the authoritative source for how to encode a message. It also moves OEM specific header type info into the oem library instead of keeping it in common.

## Changes

**Header type plumbing**

- `MessageDefinition` gains `headerType` (string, from the DB) and `eMessageType` (cached int).
- `MessageDatabase::RegisterHeaderTypeMapping` — per-message-family name→int registry, mirroring `RegisterAlignmentFunction`. OEM registers at static init alongside its alignment function.
- `eMessageType` is resolved on JSON load and re-resolved by `AppendMessages`/`Merge`/`SetMessageFamily`, since the numbering is family-specific.
- `oem::HEADER_TYPES` (`STANDARD`/`SHORT`/`STANDARD_ENCRYPTED`/`NOVATELX`) with `GetHeaderType`, `HeaderTypeToString`, `IsSupportedHeaderType`.

**Unsupported types rejected**

- `STANDARD_ENCRYPTED` and `NOVATELX` reflect header types that are not currently supported by EDIE and are not included in our current database. They are an official part of the database schema for 1.1.0 so they are parsed but raise errors if used in a decoder or encoder.

**DB schema versioning**

- New `common/version_string.hpp` (`VersionId`, `parse_version`).
- `meta` is now mandatory; schema must be ≥1.0.0 and <2.0.0. Schemas below 1.1.0 load with a warning that values may be defaulted. In particular every message will be treated as though it has a "STANDARD" header.

**Database** — regenerated at schema 1.1.0: `headerType` on every message, additional enumerators, NMEA log definition removed.

**Cleanup** — removed `hasShortHeader`; dropped the unused `stEmbeddedMetaData_` parameter from `RxConfigHandler`'s four format-specific encoders; removed dead length-computation in `EncodeBinary`.

## Breaking changes

- `novatel::edie::HEADER_FORMAT` → `DECODE_FORMAT`. `SHORT_BINARY`, `SHORT_ASCII`, `SHORT_ABB_ASCII`, `PROPRIETARY_BINARY` are gone from it; `MetaDataBase::eFormat` and all `Filter` overloads change type. `IsShortHeaderFormat` removed.
- **Python:** `HEADER_FORMAT` is renamed to `DECODE_FORMAT`, and `SHORT_ASCII`/`SHORT_BINARY`/`SHORT_ABB_ASCII`/`PROPRIETARY_BINARY` no longer exist on it. Short ASCII/binary logs report `DECODE_FORMAT.ASCII`/`BINARY`.
- `Encoder::EncodeHeader` / `EncodeBody` take `HEADER_TYPES` in place of the short-header `bool`.
- Databases must carry a `meta` block with a supported `version`.

## Tests

`test_oem_message_smoke.py` adds short-vs-standard header coverage: binary header length per message type (RAWIMUSX vs BESTPOS), short binary header field-by-field, short ASCII header text, and a short-binary round trip. C++ and Python format tests updated for the enum change.


src/decoders/oem/src/header_decoder.cpp
```diff
@@ -304,8 +304,9 @@ STATUS HeaderDecoder::Decode(const unsigned char* pucLogBuf_, IntermediateHeader
         default: return HEADER_FORMAT::UNKNOWN;
         }
     }();
+    stMetaData_.eFormat = headerFormatToDecodeFormat(eHeaderFormat);
 
-    switch (stMetaData_.eFormat)
+    switch (eHeaderFormat)
```
seanovens, line 309: Is this the only spot where we're still using `HEADER_FORMAT`? If so, is it possible to get rid of `HEADER_FORMAT` completely and just switch on the decode format (with subcases for short ASCII/Binary)?

include/novatel_edie/decoders/oem/common.hpp
```diff
@@ -71,6 +75,132 @@ constexpr uint16_t OEM4_SHORT_BINARY_SYNC_LENGTH = 3;
...
+//! definitions belonging to the "OEM" message family. The numbering is part of
+//! the mapping registered with MessageDatabase::RegisterHeaderTypeMapping, so the
+//! two must be kept in step; use GetHeaderType() to read a definition rather than
+//! comparing eMessageType directly.
+//!
+//! STANDARD is deliberately DEFAULT_HEADER_TYPE (0), making it the fallback for a
+//! definition whose header type could not be resolved. New header types must
+//! therefore be appended rather than inserted at the front.
+//-----------------------------------------------------------------------
+enum class HEADER_TYPES
```
seanovens, line 131: Can we call this `HEADER_TYPE`?

seanovens (approved):
Nice changes - the `Encode` interface feels much cleaner now

PR #260, opened 2026-08-17
Generalized field array access

seanovens:
## Summary

- Field arrays can now be accessed through the `FieldArray` type, which transparently handles type dispatch on the underlying storage (either `FlatFieldArray` or `CompositeFieldArray`)
- Adds the `GetFieldValueByName` function to `CompositeField` and `FieldArrayRecordView`, which looks up a field's definition and then retrieves its value
- Updates `README.md` with more detailed instructions about accessing field values

## More details

Previously, accessing a field array required either compile-time knowledge of the underlying storage type (`FlatFieldArray` or `CompositeFieldArray`), e.g.

```
obsArray = stMessage.GetFieldValue<FlatFieldArray>(*obsDef);
// OR
obsArray = stMessage.GetFieldValue<CompositeFieldArray>(*obsDef);
```

or using `std::visit`, e.g.

```
std::visit([](auto& arr) {
    using ArrT = std::decay_t<decltype(arr)>;
    if constexpr (std::is_same_v<ArrT, FlatFieldArray> || std::is_same_v<ArrT, CompositeFieldArray>)
    { ... }
}, stMessage.GetVarFields()[obsDef->index]);
```

The new `FieldArray` type gives a consistent way to access either kind of field array transparently:

```
obsArray = stMessage.GetFieldValue<FieldArray>(*obsDef);
```

### Access by field name

The new `GetFieldValueByName` method can be used to obtain a field's value directly using its name. This is less computationally efficient than caching a definition and reusing it for subsequent lookups, as the field's definition must be found via linear search through `FieldInfo::messageOrderedFields`. I've added more detail to the readme to explain how to access field values in more detail and to recommend by-definition lookup when possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,21 @@ while (eStatus != STATUS::STREAM_EMPTY)
 
 The same accesses can use `stMessage.GetFieldValueByName<FieldArray>("obs")` and `obs.GetFieldValueByName<float>("C_No")`, but that performs a definition search for every access and is slower. Advanced users can request `CompositeFieldArray` or `FlatFieldArray` directly when the concrete storage representation is known. `FieldArray` is preferred when code should work with either representation.
 
+> [!WARNING]
+> The `CompositeField` that owns a message's data must remain alive for as long as any field values obtained from that message are being used.
+> For example, the following code is unsafe:
+
+```
+FieldArray obsArr;
+if (...)
+{
+    CompositeField stMessage;
+    ...
+    obsArr = stMessage.GetFieldValueByName<FieldArray>("obs");
+} // Lifetime of stMessage ends after this block
+obsArr.GetFieldValueByName<float>("C_No"); // BAD - undefined behaviour!
+```
+
 #### Copy to generated struct (specialized use cases)
 
 If your input data is in the flattened binary format then, after decoding a message's header, you may copy it to the appropriate struct and access its fields as member variables.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -251,69 +251,122 @@ Run the resulting executable with the following command: `rxconfig_handler.exe <
...
     }
 }
-``````
+```
+
+The same accesses can use `stMessage.GetFieldValueByName<FieldArray>("obs")` and `obs.GetFieldValueByName<float>("C_No")`, but that performs a definition search for every access and is slower. Advanced users can request `CompositeFieldArray` or `FlatFieldArray` directly when the concrete storage representation is known. `FieldArray` is preferred when code should work with either representation.
+
+> [!WARNING]
+> The `CompositeField` that owns a message's data must remain alive for as long as any field values obtained from that message are being used.
+> For example, the following code is unsafe:
```
rldoris, line 359: Can we provide the "correct" way to do this, even if we've already provided it earlier. We should reinforce the way to handle this correctly.

README.md
```diff
@@ -326,7 +379,7 @@ This will generate a header file called `novatel_message_definitions.hpp` in you
 - Direct access to fields via member variables (very fast)
 
 **⚠️ Disadvantages:**
-- Only works with the flattened binary format
+- Only works with the flattened binary format (if your data is in a different format, you must first encode it to flattened binary)
```
rldoris: Provide the actual ENUM name.

rldoris (approved):

rldoris (approved):

PR #261, opened 2026-08-19
Expose MessageCounts in the Parser and FileParsers

rldoris:
# Expose MessageCounts in the Parser and FileParsers

## Summary

Exposes the per-message decode counts to callers of the OEM decoders.

The header decoder already counted messages, but the map was private. This PR moves the counting into a reusable `MessageCountsTracker` and exposes the counts through `HeaderDecoder`, `Parser` and `FileParser`, in both C++ and Python.

`MessageCountsTracker` takes the key type and its hash as template parameters. Decoders that identify a message in different ways therefore share one counting implementation. The OEM header decoder keys on `(message ID, format, sibling ID)`. A decoder with a simpler notion of message identity can key on the message ID alone.

## C++ API

- `novatel::edie::MessageCountsTracker<Key, Hash>` in `decoders/common` — `Increment()`, `GetCounts()`, `Reset()`. Header-only.
- `HeaderDecoder::GetMessageCounts()` and `ResetMessageCounts()`. `MessageCountsKey`, `MessageCountsKeyHash` and `MessageCountsMap` are now public.
- `Parser::GetMessageCounts()` and `ResetMessageCounts()`.
- `FileParserBase::GetMessageCounts()` and `ResetMessageCounts()`.

`MessageCountsKey` is `std::tuple<uint16_t, HEADER_FORMAT, uint8_t>`. The fields are in the order that `Filter::IncludeMessageId(id, format, source)` takes them, so the codebase has one canonical shape for this triple.

## Behavior

- The counts are cumulative.
- The format is part of the key. One log in ASCII and in binary gives two counts. The sibling ID is part of the key as well.
- A message whose ID the decoder cannot resolve is not counted.
- `FileParserBase::Reset()` clears the counts. `SetStream()` calls `Reset()`, so a new stream or a new pass over the same file starts from 0.
- The tracker is not thread-safe. `Decode()` is `const` and writes the counts through a `mutable` member, so two threads that share one decoder race on the map.

### Breaking change for other FileParserBase users

`FileParserBase::Reset()` now calls `clMyParser.ResetMessageCounts()`. Every `ParserT` used with `FileParserBase` must provide that method.

## Python API

`Parser` and `FileParser` each gain a `message_counts` property and a `reset_message_counts()` method. `FileParser.reset()` clears the counts as well.

    parser.message_counts        # {(42, HEADER_FORMAT.ASCII, 0): 2}
    parser.reset_message_counts()

The counts map binds straight through to the C++ container, with no conversion layer. The key is the same `(message_id, format, source)` tuple in both languages, and the same shape `Filter.message_ids` uses, so the two interoperate directly:

    set(parser.message_counts) == set(filter.message_ids)

The dictionary comes from a `std::unordered_map`, so key order is arbitrary. Callers that want a stable order use `sorted(counts.items())`.

## HEADER_FORMAT is now uint8_t

`enum class HEADER_FORMAT` gains an explicit `uint8_t` underlying type, matching `TIME_STATUS`. The values run from 1 to 11, so unsigned costs nothing and avoids sign extension when the enum is cast into a packed key.

This changes struct layout. `MetaDataBase` declares `bool bResponse` immediately before `eFormat`, so the enum now packs into the following byte and every later member shifts — `usWeek` moves from offset 8 to offset 2. `FramerBase`'s inner struct shifts the same way. Total sizes are unchanged, because the `double` alignment and `std::string` dominate, but the offsets are not. Anything linking a prebuilt EDIE must be rebuilt.

## Hash fix

`MessageCountsKeyHash` combined three `std::hash` results with 1-bit and 2-bit shifts. The shifts are bit shifts, not field widths, so the three fields overlapped: the sibling ID sat inside the bits of the message ID, and the format sat inside them as well. Two examples, both in `BINARY` format:

    (id 1, BINARY, sib 0)              ->  2 ^ 2 ^ 0  =  0
    (id 3, BINARY, sib 1)              ->  2 ^ 6 ^ 4  =  0
    (id 42, BINARY, sib 0)             ->  2 ^ 84     =  86
    (id 41, PROPRIETARY_BINARY, sib 0) ->  4 ^ 82     =  86

The effect depended on the standard library. MSVC hashes integers with FNV-1a, so the Windows builds were unaffected. libstdc++ and libc++ return integers unchanged, so the GCC and Clang builds lost most of the key space: with the 606 message IDs in `database/database.json` across the 10 formats, 6060 keys produced 2829 distinct hash values, and 24240 keys (siblings 0 to 3) produced 3344.

The hash now packs each field into its own bits: sibling ID in bits 0 to 7, message ID in bits 8 to 23, format above them. Every key gets its own hash value on every platform, and the OEM lookups no longer call `std::hash` three times.

## Tests

20 new C++ tests and 9 new Python tests:

- `src/decoders/common/test/message_counts_tracker_unit_test.cpp` — the template alone, with both a tuple key and a plain integer key.
- `src/decoders/oem/test/message_counts_test.cpp` — the hash bit layout, an exhaustive uniqueness check over 132,000 keys, and the counts through `HeaderDecoder`, `Parser` and `FileParser`.
- `python/test/test_parser.py` and `python/test/test_file_parser.py` — the new property and method, and the tuple compatibility with `Filter.message_ids`.

Full suites pass: 404 C++ tests, and 1041 Python tests. Both were run on Windows with MSVC. The Python bindings were built with `BUILD_PYTHON=1` and the generated stub reads `dict[tuple[int, common_bindings.HEADER_FORMAT, int], int]`.


include/novatel_edie/decoders/oem/header_decoder.hpp
```diff
@@ -40,29 +44,42 @@ namespace novatel::edie::oem {
...
+    //! two different hash values.
     struct MessageCountsKeyHash
     {
         std::size_t operator()(const MessageCountsKey& key) const
         {
-            auto h1 = std::hash<int>{}(static_cast<int>(std::get<0>(key)));
-            auto h2 = std::hash<uint16_t>{}(std::get<1>(key));
-            auto h3 = std::hash<uint8_t>{}(std::get<2>(key));
-            return h1 ^ (h2 << 1) ^ (h3 << 2);
+            return (static_cast<std::size_t>(std::get<0>(key)) << 24) | (static_cast<std::size_t>(std::get<1>(key)) << 8) |
```
riley-kinahan: Total nitpick but HEADER_FORMAT is technically a 32 bit int so it gets cut off by this hash function on a 32 bit architecture. Either use `uint64_t` or change the `HEADER_FORMAT` definition to 

```
enum class HEADER_FORMAT : uint8_t
{
    UNKNOWN = 1,
    BINARY,
    SHORT_BINARY,
    PROPRIETARY_BINARY,
    ASCII,
    SHORT_ASCII,
    ABB_ASCII,
    NMEA,
    JSON,
    SHORT_ABB_ASCII,
    ALL // Used in filters to indicate all filter types : all new enums should be added before this value
};
```

seanovens:
> libstdc++ and libc++ return integers unchanged

Wow, did not know this - oops!

include/novatel_edie/decoders/oem/header_decoder.hpp
rldoris: Updated

include/novatel_edie/decoders/oem/header_decoder.hpp
```diff
@@ -40,29 +44,42 @@ namespace novatel::edie::oem {
 //============================================================================
 class HeaderDecoder
 {
-  private:
+  public:
     //! \brief Type alias for message counts key.
-    //! \details Tuple containing: format (ASCII, binary, etc.), message ID (uint16_t), sibling ID (uint8_t).
+    //! \details Tuple containing: format (ASCII, binary, and more), message ID (uint16_t), sibling ID (uint8_t).
+    //! The same message in two formats has two keys, and therefore two counts.
     using MessageCountsKey = std::tuple<HEADER_FORMAT, uint16_t, uint8_t>;
```
seanovens: Minor comment/question: if we reordered `MessageCountsKey` to be `std::tuple<uint16_t, HEADER_FORMAT, uint8_t>`, could we avoid having the `PyMessageCountsMap` shim and just bind directly to the C++ message counts map (while maintaining the symmetry between the map keys and `filter.message_ids`)?

seanovens (approved):

PR #262, opened 2026-08-31
Add Header Types

riley-kinahan:
Previously we relied on the format of a decoded message to determine whether to encode it with a standard or short header. This meant that directly constructed message objects would be encoded with an incorrect header type unless metadata was deliberately constructed with this in mind.

This PR adds a "headerType" field to the database which acts as the authoritative source for how to encode a message. It also moves OEM specific header type info into the oem library instead of keeping it in common.

## Changes

**Header type plumbing**

- `MessageDefinition` gains `headerType` (string, from the DB) and `eMessageType` (cached int).
- `MessageDatabase::RegisterHeaderTypeMapping` — per-message-family name→int registry, mirroring `RegisterAlignmentFunction`. OEM registers at static init alongside its alignment function.
- `eMessageType` is resolved on JSON load and re-resolved by `AppendMessages`/`Merge`/`SetMessageFamily`, since the numbering is family-specific.
- `oem::HEADER_TYPE` (`STANDARD`/`SHORT`/`STANDARD_ENCRYPTED`/`NOVATELX`) with `GetHeaderType`, `HeaderTypeToString`, `IsSupportedHeaderType`.

**Unsupported types rejected**

- `STANDARD_ENCRYPTED` and `NOVATELX` reflect header types that are not currently supported by EDIE and are not included in our current database. They are an official part of the database schema for 1.1.0 so they are parsed but raise errors if used in a decoder or encoder.

**DB schema versioning**

- New `common/version_string.hpp` (`VersionId`, `parse_version`).
- `meta` is now mandatory; schema must be ≥1.0.0 and <2.0.0. Schemas below 1.1.0 load with a warning that values may be defaulted. In particular every message will be treated as though it has a "STANDARD" header.

**Database** — regenerated at schema 1.1.0: `headerType` on every message, additional enumerators, NMEA log definition removed.

**Cleanup** — removed `hasShortHeader`; dropped the unused `stEmbeddedMetaData_` parameter from `RxConfigHandler`'s four format-specific encoders; removed dead length-computation in `EncodeBinary`.

## Breaking changes

- `novatel::edie::HEADER_FORMAT` → `DECODE_FORMAT`. `SHORT_BINARY`, `SHORT_ASCII`, `SHORT_ABB_ASCII`, `PROPRIETARY_BINARY` are gone from it; `MetaDataBase::eFormat` and all `Filter` overloads change type. `IsShortHeaderFormat` removed.
- **Python:** `HEADER_FORMAT` is renamed to `DECODE_FORMAT` with backwards compatibility alias,  but `SHORT_ASCII`/`SHORT_BINARY`/`SHORT_ABB_ASCII`/`PROPRIETARY_BINARY` no longer exist on it. Short ASCII/binary logs report `DECODE_FORMAT.ASCII`/`BINARY`.
- `Encoder::EncodeHeader` / `EncodeBody` take `HEADER_TYPES` in place of the short-header `bool`.
- Databases must carry a `meta` block with a supported `version`.

## Tests

`test_oem_message_smoke.py` adds short-vs-standard header coverage: binary header length per message type (RAWIMUSX vs BESTPOS), short binary header field-by-field, short ASCII header text, and a short-binary round trip. C++ and Python format tests updated for the enum change.


src/decoders/oem/src/message_decoder.cpp
```diff
@@ -37,10 +37,12 @@
...
 using namespace novatel::edie::oem;
 
-// Register the OEM alignment function with the MessageDatabase at static initialization time
+// Register the OEM alignment function and header type mapping with the MessageDatabase at static
+// initialization time, so that both are in place before any database is read.
 namespace {
-const bool kRegisteredOemAlignment = [] {
+const bool kRegisteredOemFamily = [] {
     novatel::edie::MessageDatabase::RegisterAlignmentFunction("OEM", novatel::edie::oem::OemAlignmentFunction);
+    novatel::edie::MessageDatabase::RegisterHeaderTypeMapping("OEM", novatel::edie::oem::OemHeaderTypeMapping());
```
seanovens, line 45: I recently noticed a flaw in my approach here: if a consumer application only uses the OEM Encoder (and not the decoder) and doesn't provide the necessary linker flags to preserve this code, it will get eliminated (and the Encoder might not work correctly). In #266 I addressed this by moving the static registration code to `include/novatel_edie/decoders/oem/common.hpp`, but I'm also open to suggestions if you have a better idea

cmake/conan_provider.cmake
seanovens, line 23: Should this be 2.25.0 now since we're switching over to `CMakeConfigDeps`? Doesn't matter for this project since the minimum is already enforced in the conanfile, but could save some effort later if this file is copy-pasted to a new project

include/novatel_edie/decoders/common/message_database.hpp
```diff
@@ -833,6 +859,21 @@ class MessageDatabase
...
+    //
+    //! \param[in] iMsgId_ The message ID.
+    //! \return A pointer to the definition, or nullptr if the ID is not in the database.
+    //----------------------------------------------------------------------------
+    [[nodiscard]] const MessageDefinition* GetMsgDefRaw(const int32_t iMsgId_) const
+    {
+        const auto it = mMessageId.find(iMsgId_);
+        return it != mMessageId.end() ? it->second.get() : nullptr;
+    }
+
```
seanovens, line 876: Out of curiosity, did this have a significant impact on any of the encoding benchmarks? If so, might be worth using this on the decoder side as well (future work)

PR #263, opened 2026-09-04
Documentation updates

riley-kinahan:
Sync up documentation with recent development work.

I also included a one-line bugfix for setting STRING values from python, and added tests that will catch any regression. Previously data wrote into the fixed length data structure when it should have gone into the variable length one.


rldoris (approved):

rldoris (changes requested):
Fix CI/CD

